PR #4, opened 2020-03-06
make usage of serialport and crc optional

Dushistov:
Summary of changes: 
- disable tests that doesn't compile
- use workspace and override to speedup compilation
- make serialport and crc optional

It is only move around PR, so public API and other things should not change,
only visible outside change is that there is no default serialport usage.

refs #1


Dushistov:
@lkolbly 

I would like disscuss and merge this and then make another PR, that summarize our discussion about proc macro usage, syntax and so on things.

ublox_serial_demo/src/main.rs
```diff
@@ -1,16 +1,20 @@
-use ublox::{Device, Position};
-use std::time::Duration;
 use chrono::prelude::*;
+use std::time::Duration;
+use ublox::{Device, Position};
 
 fn main() {
```
lkolbly: This is better suited as an example, perhaps `examples/serial.rs`? (or I guess, `examples/serial.rs` under the `ublox` crate)

Cargo.toml
```diff
@@ -1,19 +1,6 @@
...
-bincode = "1.2.1"
-chrono = "0.4"
-crc = "1.8.1"
-syn = "1.0.14"
-ublox_derive = { path = "ublox_derive/" }
-num-traits = "0.2"
-num-derive = "0.2"
+[patch.'crates-io']
+ublox_derive = { path = "ublox_derive" }
+ublox = { path = "ublox" }
```
lkolbly: Add trailing newline

ublox/src/ubx_packets.rs
```diff
@@ -168,45 +121,21 @@ pub struct NavPosLLH {
...
-            speed: 0.0,
-            heading: 0.0,
+            lon: packet.lon as f32 / 10_000_000.0,
+            lat: packet.lat as f32 / 10_000_000.0,
+            alt: packet.height_msl as f32 / 1000.0,
         }
     }
 }
 
-#[ubx_packet]
```
lkolbly: > I would like disscuss and merge this and then make another PR, that summarize our discussion about proc macro usage, syntax and so on things.

It may make sense to leave the `ubx_packet` macro usages in for this PR, if in the end we decide to go with something similar no sense in removing them now.

ublox_serial_demo/src/main.rs
Dushistov: Done

Cargo.toml
Dushistov: Done

ublox/src/ubx_packets.rs
Dushistov: Done

PR #5, opened 2020-03-09
CI fixes

Dushistov:
setup "script" by hands, to check:
- how tests works with debug build (arithmetic overflow is only
checked in debug release)
- check that documentations builds without errors
- check build with feature = "serial"
- check build of examples


Dushistov:
I don't use matrix to speedup CI run (to not launch VM for each feature). In fact it would be nice to migrate to github actions to speedup build and also to build for windows (and mac) just in case.

PR #7, opened 2020-03-11
Use own proc macro to generate code to deal with packets

### LINKED ISSUE #1: separate protocol and implementation?

Dushistov:
Features:
- removes unnecessary dependicies
- more type safe code
- faster parser: 3x
fixes #1

@lkolbly 

Problems I found during conversation to usage of proc macro that I would like to disscuss:

1. I replaced the end of `NavPosVelTime`
before:
```rust
    pos_dop: u16,
    reserved2: u16,
    pub reserved3: u32,
```
after:
```rust
    pos_dop: u16,
    reserved1: [u8; 6],
    heading_of_vehicle: i32,
    magnetic_declination: i16,
    magnetic_declination_accuracy: u16,
```
the last version is match documentation, while previous one not,
not sure how this works before

2. `NavPosVelTime` uses mm/s as speed units while `NavVelNed` uses cm/s
    according to documentation, but previous code for some reason treats `NavVelNed`
   as mm instead of cm, is it problem with documentation or with code?
   And side note `NavStatus` have similar fields with `NavPosVelTime` , but only one bit representation match according to documentation, all other fields has different format, though the same name and the same purpose.

3. Why negative nanoseconds values converted to 0, may be negative also valid value, or why they use signed integer instead of unsigned one?

4. May be use chrono function to validate date instead of convert it without check?

What missed:
- testing on real device (I am goging to do it tomorrow)
-  full conversation of code with features=serial , I only partly converted it,
   because of how you plan to use AlpSrv is not clear for me,
   and looks like AlpSrv is removed in u-blox8 (https://www.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_(UBX-13003221)_Public.pdf)
  But I convert what I can, and you can see how code improved, at now you don't need to use magic numbers for class/message id.

And comment from previous review: 
>  The create_packet signature looks good to me (would it make sense to call it serialize

This is against idea that `Builder` is just the way to create packet, so you can not serialize builder, there is no sense in this operation, but you can create packet from builder. That's why I don't call it serialization.


lkolbly:
> full conversation of code with features=serial

Don't spend too much effort on this part - I want to totally rewrite how the serial library parts work, so if anything gets sticky with that, feel free to just comment it out entirely.

I'll look more at this later when I get a sec, but in general where the documentation and the code disagree I trust the documentation more.

Dushistov:
Fixed issue with `UbxPacketRequest` that I found during 
testing with real device.

> so if anything gets sticky with that, feel free to just comment it out entirely

I did almost that.

.travis.yml
```diff
@@ -9,6 +9,9 @@ matrix:
   allow_failures:
     - rust: nightly
   fast_finish: true
+before_script:
+  - rustup component add rustfmt-preview
+  - which rustfmt
```
lkolbly: What does this do?

ublox/src/ubx_packets/packets.rs
```diff
@@ -0,0 +1,718 @@
...
+        }
+    }
+}
+
+/// Receiver/Software Version
+#[ubx_packet_recv]
+#[ubx(class = 0x0a, id = 0x04)]
+struct MonVer {
+    #[ubx(map_type = &str, may_failed, from = mon_ver::convert_to_str_unchecked,
+          is_valid = mon_ver::is_cstr_valid, get_as_ref)]
```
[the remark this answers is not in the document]
Dushistov: - `is_valid` used to distinguish packets with the same class and the same message id,
    like CfgPrt(Uart|Spi|Usb)
- it will be very inconvenient to use it, imagine you got PacketX with several invalid fields,
   You need manually check all fields like:
       ```rust
       match (pack_ref.field1(), pack_ref.field2(), pack_ref.field3()) {
          (Some(f1), Some(f2), Some(f3)) => ok_we_got_what_we_want,
          None => return err,
       }
      ```
      and `Option` means that there is no such field, while there is such field, but packet broken, and why user have to manually handle such errors?

In case of MonVer, missed 0 char, extenstion length not match 30 * N, means that something wrong with channel, or something like bit-flip happens in memory of GPS receiver, so all data inside of packet may be not valid, TCP checksum used in UBX protocol is not that [great](https://stackoverflow.com/questions/3830206/can-a-tcp-checksum-fail-to-detect-an-error-if-yes-how-is-this-dealt-with)

ublox/src/ubx_packets/packets.rs
```diff
@@ -0,0 +1,718 @@
...
+        self.ecef_x_or_lat = (pos.lat * 10_000_000.0) as i32;
+        self.ecef_y_or_lon = (pos.lon * 10_000_000.0) as i32;
+        self.ecef_z_or_alt = (pos.alt * 100.0) as i32; // Height is in centimeters, here
+        self.flags |= (1 << 0) | (1 << 5);
+        self
+    }
+
+    pub fn set_time(mut self, tm: DateTime<Utc>) -> Self {
+        self.week_or_ym = (match tm.year_ce() {
+            (true, yr) => yr - 2000,
```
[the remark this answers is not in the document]
Dushistov: Hm.. have no idea, I just move already existing code around: https://github.com/lkolbly/ublox/blob/898a6381ccee2612175cfe04a97ce0f29b033707/ublox/src/ubx_packets.rs#L329 

I plan to add use chrono to validate input packets in the next patch series,
but I don't plan yet use AidIni, so have no idea what is this code about.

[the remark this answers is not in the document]
lkolbly: Alright, you can leave it how it is then.

ublox/src/ubx_packets/packets.rs
```diff
@@ -0,0 +1,718 @@
...
+    time_accuracy: u32,
+    nanosecond: i32,
+    /// GNSS fix Type
+    #[ubx(map_type = GpsFix)]
+    fix_type: u8,
+    flags: u8,
+    flags2: u8,
+    num_satellites: u8,
+    #[ubx(map_type = f64, scale = 1e-7, alias = lon_degrees)]
+    lon: i32,
```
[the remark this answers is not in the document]
Dushistov: Because of this field doesn't contain degrees. It contains degree multiply on 1e7.
If we add (in the future) interface to get "raw" fields, it would be good to not deceive user,
and distinguish between degrees/f64 and raw field where user need handle by him/her self.

The plan actually have the same units as getter everywhere if possible:
meters, meters per second, degrees. But while meters and seconds are obvious choice from [International System of Units](https://en.wikipedia.org/wiki/International_System_of_Units). 

Usage of degrees not so obvious, because of all functions from stdlib that work with angles works with radians in C/Rust and other standard libraries (like cos/sin/tan and so on),
so for degree I add such "comment" via method name, may be better return something like `struct Degree<T>(pub T)`.

ublox/src/ubx_packets.rs
```diff
@@ -1,517 +1,128 @@
...
-    pub pos_dop: u16,
-    pub reserved2: u16,
-    pub reserved3: u32,
+/// Abstraction for buffer creation/reallocation
+/// to storing packet
+pub trait MemWriter {
+    type Error: std::error::Error;
+    /// make sure that we have at least `len` bytes for writing
+    fn reserve_allocate(&mut self, len: usize) -> Result<(), MemWriterError<Self::Error>>;
+    fn write(&mut self, buf: &[u8]) -> Result<(), MemWriterError<Self::Error>>;
```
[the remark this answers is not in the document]
Dushistov: > Will this trait work for e.g. [u8; 256]? 

If there is  packet with such field type (`[u8; 256]`) it should, in theory maximum "write size" can be up to `u16::max_value() `,
but currently we have no such packet in `packet.rs`

> It looks like in order to implement it we would need to keep track of the offset/number of bytes written.

Yep, there is need one variable, minimum `u16` for current offset.
Alternatively we can allocate full packet on stack and only then call `MemWriter::write`,
while we currently call `MemWriter::write` for each field.
But that would require "huge" allocation on stack, of course I mean huge in compare with one `u16` variable. And I suppose in general compiler can "inline" all write calls for cases like this:

```
struct DmaBufWriter {
   cur_write_off: u16,
   dma_buf: *mut u8,
}
impl MemWriter for DmaBufWriter {
   fn reserve_allocate(&mut self) { self.dma_buf = dma_alloc_coherent(); }
   fn write(&mut self, buf: &[u8]) { 
     memcpy(self.dma_buf.offset(self.cur_off), buf.as_ptr(), buf.len()); 
     self.cur_off += buf.len();
  }
}
```

.travis.yml
Dushistov: rustfmt used for tests, and rustfmt is not installed by default, so these lines installed it.

ublox/src/parser.rs
```diff
@@ -0,0 +1,121 @@
+use crate::{
+    error::ParserError,
+    ubx_packets::{match_packet, ubx_checksum, PacketRef, SYNC_CHAR_1, SYNC_CHAR_2},
+};
+
+/// Some big number, TODO: need validation against all known packets
+const MAX_PACK_LEN: usize = 1022;
```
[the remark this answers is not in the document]
Dushistov: This is constant from real project to protect against "invalid" byte sequences. In order not to go further and not wait many bytes before we at last can verify checksum and find out that this is not real packet, but just strange byte sequence.

In theory yes, 2^16, but there are no such packets exists, so valid constant I suppose should be max against all fixed_packet_len and something up limit for packets without fixed length, like `MonVer`. 1022 greater that any "fixed" length of known packets and (1022-40)/30 is big enough. I suppose in future we can automatically calculate right size during handling of `define_recv_packets!`

ublox/src/ubx_packets/packets.rs
```diff
@@ -0,0 +1,718 @@
...
+#[ubx_packet_send]
+#[ubx(class = 6, id = 1, fixed_payload_len = 8)]
+struct CfgMsg8 {
+    msg_class: u8,
+    msg_id: u8,
+    /// Send rate on I/O Port (6 Ports)
+    rates: [u8; 6],
+}
+
+impl CfgMsg8Builder {
```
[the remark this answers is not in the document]
Dushistov: > CfgMsgRateBuilder

This name derived from packet name: CfgMsg8, so you suggest call "packet" with `CfgMsgRate` name?
But this would be against naming scheme - use the same names for packets as in ubx protocol documentation.

ublox_derive/src/output.rs
```diff
@@ -0,0 +1,562 @@
...
+                } else {
+                    Err(ParserError::InvalidPacketLen{ packet: #pack_name, expect: min, got })
+                }
+            }
+        }
+    };
+
+    quote! {
+        #[doc = #struct_comment]
+        #[doc = "It is just reference to internal parser's buffer"]
```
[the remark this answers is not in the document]
Dushistov: > Conceivably, an end user could use one of these to reference their own internal parser buffer, right?

No, definition looks like this `struct TestRef<'a>(&'a [u8])` (from ublox_derive/src/tests.rs), as you can see content of `TestRef` is private, so user can not construct it by (him|her)self  in his/her  "internal parser".

Because of `PacketRef` is designed in invariant that it created against "valid buffer",
and our parser can guarantee this, while about other "internal parser" it is absolutely unknown.

ublox/tests/parser_tests.rs
```diff
@@ -0,0 +1,157 @@
...
+    assert_eq!(
+        my_vec![Ok((6, 1))],
+        extract_ack_ack(parser.consume(&garbage_before)),
+        "garbage before1"
+    );
+}
+
+#[test]
+#[ignore]
+fn test_parse_big_file() {
```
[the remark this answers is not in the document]
Dushistov: Actually this is the only test I trust. It is big enough, I validate number of packets in each "class" using three different parsers (during benchmarking), I turn on NMEA output also, so this "big file" contains "random" between packets. So this is actually the only test that really validate parsing and all generated code.

What about I would replace these lines:
```
    assert_eq!(0, wrong_chksum);
    assert_eq!(0, other_errors);
    assert_eq!(38291, nav_pos_llh);
    assert_eq!(38291, nav_stat);
    assert_eq!(120723, unknown);
    assert_eq!(1, ack_ack);
```
with data from `format!("{}.metadata", env::var("UBX_BIG_LOG_PATH").unwrap())`,
so you can also can run this test with your own big enough log file, via something like:
```sh
UBX_BIG_LOG_PATH=/var/tmp/gps.bin time cargo test --release test_parse_big_file -- --ignored --nocapture
```
?

ublox/src/parser.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+
+impl<'a> ParserIter<'a> {
+    /// Analog of `core::iter::Iterator::next`, should be switched to
+    /// trait implmentation after merge of https://github.com/rust-lang/rust/issues/44265
+    pub fn next(&mut self) -> Option<Result<PacketRef, ParserError>> {
+        while self.off < self.buf.len() {
+            let data = &self.buf[self.off..];
+            let pos = match data.iter().position(|x| *x == SYNC_CHAR_1) {
```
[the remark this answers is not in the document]
Dushistov: Modifying `self.off` is bad idea, but it is good idea to create local variable with `&data[pos`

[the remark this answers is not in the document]
lkolbly: Why is modifying `self.off` a bad idea? Don't we do it anyway below?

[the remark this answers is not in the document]
Dushistov: > Why is modifying self.off a bad idea? Don't we do it anyway below?

My supposition, 99% of time we will handle continuous stream of valid packets.
In that case it would be like this (with `self.off += pos` modification as you suggest):
```
let pos = 0;//99% cases

self.off += 0; // actually `self.off += pos` , but in 99% pos == 0`
if not enough data {
   return None;
}
self.off += 6 + pack_len + 2;
```

you see, it would be the same code with `self.off +pos` at start and without if we get full packet, but if would be unnecessary
`self.off +=0` every time for partly received packet.

So if user pass buffer with length 1 to `consume` (as for example [here](https://github.com/lkolbly/ublox/blob/567263b5011a7367236de1f6f0ce5e6528d99d76/ublox/src/serialport.rs#L393) and in my own code, untill I read protocol and changed buffer length to 8) we get `28/*NavPosLlh payload*/ + 6 + 2` calls to `next` and only one of them will really update  `self.off`, while the others would only:
```
self.off += 0;
```
and compiler can not optimize this, because of it calculate `pos` every time.

ublox/tests/parser_tests.rs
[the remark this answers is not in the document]
lkolbly: Alright, that would work for me. Add a comment somewhere saying how to run it, and maybe move it to a different test file (binary_dump_test.rs or something).

ublox/src/parser.rs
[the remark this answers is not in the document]
lkolbly: Okay, I'm fine with deferring this. Why 1022, and not a round number like 1024?

ublox/src/ubx_packets/packets.rs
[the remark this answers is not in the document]
lkolbly: I guess really the question is, where are the names `CfgMsg8` and `CfgMsg3` coming from? Should they be called `CfgMsgSinglePort` and `CfgMsgAllPorts`?

ublox/src/parser.rs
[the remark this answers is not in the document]
Dushistov: > Why 1022

I suppose 1022 is size of packet, have no idea what packet. I added automatic calculation of maximum packet size.

ublox/src/ubx_packets/packets.rs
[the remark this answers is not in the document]
Dushistov: > where are the names CfgMsg8 and CfgMsg3

From payload length, it should be unique. So I used it as suffix. But Single/AllPort looks better, and I renamed them.

ublox_derive/src/input.rs
```diff
@@ -0,0 +1,643 @@
...
+                    if p.is_ident("from") {
+                        from_fn = Some(UbxTypeFromFn::From);
+                    } else if p.is_ident("into_raw") {
+                        into_fn = Some(UbxTypeIntoFn::Raw);
+                    } else if p.is_ident("from_unchecked") {
+                        from_fn = Some(UbxTypeFromFn::FromUnchecked);
+                    } else if p.is_ident("rest_reserved") {
+                        rest_handling = Some(UbxEnumRestHandling::Reserved);
+                    } else if p.is_ident("rest_error") {
+                        rest_handling = Some(UbxEnumRestHandling::ErrorProne);
```
lkolbly: We should check that they don't specify both rest_reserved and rest_error (same for the `from_fn` fields)

ublox_derive/src/input.rs
```diff
@@ -0,0 +1,643 @@
...
+    };
+
+    let mut consts = Vec::with_capacity(ast.items.len());
+    for item in ast.items {
+        consts.push(BitFlagsMacroItem {
+            attrs: item.attrs,
+            name: item.name,
+            value: item.value.base10_parse()?,
+        });
+    }
```
lkolbly: Can you do this using `ast.items.iter().map(...).collect()`?

ublox_derive/src/input.rs
```diff
@@ -0,0 +1,643 @@
...
+        return Err(Error::new(
+            meta_sp,
+            "you should use rest_error with from_unchecked",
+        ));
+    }
+
+    Ok((from_fn, into_fn, rest_handling))
+}
+
+fn parse_ubx_attr(attrs: &[Attribute], struct_name: &Ident) -> syn::Result<PackHeader> {
```
lkolbly: `parse_ubx_attr` is only used for the struct parsing. `input.rs` is pretty big, should it be split into different files for interpreting structs and extending enums?

ublox_derive/src/output.rs
```diff
@@ -0,0 +1,572 @@
...
+        }
+        #from
+        #into
+    })
+}
+
+pub fn generate_code_for_parse(recv_packs: &RecvPackets) -> TokenStream {
+    let mut pack_enum_variants = vec![];
+    let mut matches = vec![];
+    let mut max_16_chain = None;
```
lkolbly: Maybe call this `max_pack_len_calculation`?

ublox_derive/src/output.rs
```diff
@@ -0,0 +1,572 @@
...
+        matches.push(quote! {
+            (#name::CLASS, #name::ID) if <#ref_name>::validate(payload).is_ok()  => {
+                Ok(PacketRef::#name(#ref_name(payload)))
+            }
+        });
+        if let Some(prev_max_16_chain) = max_16_chain {
+            max_16_chain = Some(quote! { max_u16(#name::MAX_PAYLOAD_LEN, #prev_max_16_chain) });
+        } else {
+            max_16_chain = Some(quote! { #name::MAX_PAYLOAD_LEN });
+        }
```
lkolbly: I think you could calculate this with a fold:
```
recv_packs.all_packets.iter().fold(quote!{ 0 }, |acc, x| {
   quote!{ max_u16(#acc, #x::MAX_PAYLOAD_LEN) }
});
```

ublox_derive/src/output.rs
```diff
@@ -0,0 +1,572 @@
...
+        let ref_name = format_ident!("{}Ref", name);
+        pack_enum_variants.push(quote! {
+            #name(#ref_name <'a>)
+        });
+
+        matches.push(quote! {
+            (#name::CLASS, #name::ID) if <#ref_name>::validate(payload).is_ok()  => {
+                Ok(PacketRef::#name(#ref_name(payload)))
+            }
+        });
```
lkolbly: I think these could each be generated from maps.

ublox_derive/src/output.rs
```diff
@@ -0,0 +1,572 @@
...
+                _ => Ok(PacketRef::Unknown(#unknown_var {
+                    payload,
+                    class,
+                    msg_id
+                })),
+            }
+        }
+
+        const fn max_u16(a: u16, b: u16) -> u16 {
+            [a, b][(a < b) as usize]
```
lkolbly: I think it would just be clearer to do:
```
if b > a { b } else { a }
```

lkolbly: Also, can we just use std::cmp::max?

ublox_derive/src/output.rs
```diff
@@ -0,0 +1,572 @@
...
+                    x.into_packet_bytes()
+                }
+            }
+        });
+
+        ret.push(quote! {
+            impl UbxPacketCreator for #payload_struct {
+                #[inline]
+                fn create_packet<T: MemWriter>(self, out: &mut T) -> Result<(), MemWriterError<T::Error>> {
+                    out.reserve_allocate(#packet_size)?;
```
lkolbly: Is the only reason we want `reserve_allocate` for performance reasons?

It occurred to me just now that it'd be weird to write `reserve_allocate` for, for example, a SpiWriter that writes directly to a spi port.

ublox_derive/src/output.rs
```diff
@@ -0,0 +1,572 @@
...
+                }
+            }
+            impl From<#payload_struct> for [u8; #packet_size] {
+                fn from(x: #payload_struct) -> Self {
+                    x.into_packet_bytes()
+                }
+            }
+        });
+
+        ret.push(quote! {
```
lkolbly: You should be able to combine these?

ublox_derive/src/output.rs
```diff
@@ -0,0 +1,572 @@
...
+    {
+        quote! { #[derive(Default)] }
+    } else {
+        quote! {}
+    };
+    let mut ret = Vec::with_capacity(4);
+    let struct_comment = &pack_descr.comment;
+    ret.push(quote! {
+        #[doc = #struct_comment]
+        #[doc = "Struct that used as \"builder\" for packet"]
```
lkolbly: Struct that _is_ used as...

ublox_derive/src/output.rs
Dushistov: I don't get it. It is impossible to use `if` in `const fn`, and `std::cmp::max` is not `const fn`.

ublox_derive/src/output.rs
Dushistov: > SpiWriter that writes directly to a spi port

I don't think that it is possible. API is designed for fill some buffer with data. For buffer reuse, for writing to some special location like DAM buffer continiously mapped to physical memory  and so on. That's why it called `MemWriter`. 

It can try to use this API to write directly to device, but in this case it would be hard to
handle errors like "try again latter" (EGAIN/EBUSY), because of you have not buffer, but you need restart from specific location. This is API not for direct device access.

ublox_derive/src/output.rs
lkolbly: Ah, I did not know that. Weird that `if` isn't allowed, but okay.

ublox_derive/src/output.rs
lkolbly: For sure, there are some devices which would be very hard to interface with. There is a [blocking SPI bus](https://github.com/rust-embedded/embedded-hal/blob/master/src/blocking/spi.rs) in the embedded hal, though, which would probably work fine.

ublox_derive/src/output.rs
Dushistov: For example you can use `if` inside C++ `constexpr` only starting from C++14, C++11 doesn't allow this. In fact `if` and `match` is implemented for `const fn`, but it is nigtly only, behind "feature gate" rust-lang/rust#49146

ublox_derive/src/output.rs
Dushistov: Obviously any unblocking API can be used in blocking way via busy loop. But again why care? 

This is API designed for filling memory buffers, not for "direct" transfer. If somebody not care about performance, and don't use DMA, so his/her CPU slow-down and starts work with SPI rate, like 10-20Mhz, then obviously that person doesn't care about performance at all, why care about that case?

Plus `reserve_allocate` is not only for performance reason, it is also for handling "out-of-memory" errors like https://doc.rust-lang.org/beta/std/vec/struct.Vec.html#method.try_reserve .

Plus what is the gain here? I see only the reason to use "direct" transfer in compare with into_packet_bytes/DmaBuf::write in increasing of code size by decreasing stack size. So basically if you use "direct" transfer you increase code size (many writes), if you use "into_packet_bytes"/DmaBuf::write you decrease code size, but increase stack or heap usage. 

And if we talk about our case, then our stack/heap usage is really small. May be this is significant for really small microcontrollers with 2K SRAM (stack/heap) and 32K NOR flash (code), but I really doubt that there is sense to use such kind of microcontrollers with GPS.

lkolbly: > why care about that case?

Ease of implementation for the end user. And of course, for the memory-sensitive - you say there's no sense in using a small MCU with a GPS module, but arduino's project hub has [several projects](https://create.arduino.cc/projecthub/projects/tags/gps) which combine GPS modules with controllers that have <2KB of ram.

> Plus reserve_allocate is not only for performance reason, it is also for handling "out-of-memory" errors

out-of-memory errors I think could be handled by failing in the write() call as well.

Either way, I'm fine with leaving it as it is, I was just thinking about the ramifications of this particular API.

Dushistov: > out-of-memory errors I think could be handled by failing in the write() call as well

In some cases no. For example `alloc::vec::Vec` it would be troublesome.  Only `Vec::try_reserve` return out-of-memory error. `Write::write` works without errors.

> but arduino's project hub 

In this case `PacketBuilder::into_packet_bytes` in form of `const fn` function would be much nicer. Of course when `rustc` allow `for` in `const fn` for `ubx_checksum`. Then instead of load chunks of data into SRAM for sending, it would be possible to send one whole packet in one call, and flash to hold const byte array with packet.

ublox_derive/src/input.rs
Dushistov: > ast.items.iter().map(...).collect()?

It is hard, because of `value: item.value.base10_parse()?`, see **?** at the end.

ublox_derive/src/output.rs
Dushistov: It would be more complicated then existing code,
something like this:

```
    let (x, y): (Vec<_>, Vec<_>) = recv_packs
        .all_packets
        .iter()
        .map(|name| {
            let ref_name = format_ident!("{}Ref", name);
            (
                quote! {
                    #name(#ref_name <'a>)
                },
                quote! {
                    (#name::CLASS, #name::ID) if <#ref_name>::validate(payload).is_ok()  => {
                        Ok(PacketRef::#name(#ref_name(payload)))
                    }
                },
            )
        })
        .unzip();
```

ublox_derive/src/input.rs
Dushistov: I suppose this is very subjective field: right size of module. I would suggest deal with it not in the frame of this PR.

ublox_derive/src/input.rs
lkolbly: I haven't tried it, but I think collect() should be able to collect over Result values, stopping on the first error: https://doc.rust-lang.org/rust-by-example/error/iter_result.html

Dushistov: > I haven't tried it, but I think collect() should be able to collect over Result values

1. This variant would be slower: rust-lang/rust#48994
2. It require more lines, and more complex code:
```rust
    let consts = ast
        .items
        .into_iter()
        .map(|item| {
            Ok(BitFlagsMacroItem {
                attrs: item.attrs,
                name: item.name,
                value: item.value.base10_parse()?,
            })
        })
        .collect::<syn::Result<Vec<_>>>()?;
```

11 lines vs 8, plus slower. Is it worth this?

lkolbly: I don't think performance really matters here, but I'm fine with dropping.

Dushistov: @lkolbly 

> I don't think performance really matters here

The funny thing that performance really matter in build script and procedural macroses,
specially performance in debug mode.

Because of:
1. Debug vs Release difference in Rust case is really huge, may be 20x, may be 100x difference
2. By default procedural macroses / build scripts compiled with the same optimization
    level as the main crate.

So if some day packet.rs will contain all packages from ubx spec, compilation of ublox
in debug mode may become slow. But good thing is that at least nightly cargo allow change optimization level for specific dependencies, so this is not as bad as before.

lkolbly: Even if packet.rs has every packet, it's still on the order of thousands of items at most - pre-allocating vs. not-pre-allocating the vector holding them will be an unnoticeable performance difference to the end user.

Dushistov: My comment was in general about performance of whole ublox_derive, not about specific `with_capacity`. If talk about this specific case **not** inlined iterators methods (because of debug) vs simple loop may create difference for thousands of packets.

PR #8, opened 2020-03-26
feature: add UBX-CFG-NAV5 message support

Dushistov:


ublox/src/ubx_packets/packets.rs
```diff
@@ -708,6 +708,240 @@ impl CfgMsgAllPortsBuilder {
...
+    fixed_alt_var: u32,
+
+    /// Minimum Elevation for a GNSS satellite to be used in NAV (deg)
+    min_elev_degrees: i8,
+
+    /// Reserved
+    dr_limit: u8,
+
+    /// Position DOP Mask to use
+    #[ubx(map_type = f32, scale = 0.1, into = ScaleBack::<f32>(10.).as_u16)]
```
lkolbly: We should be consistent about using f64 (or f32, but my preferences lean toward f64).

ublox/src/ubx_packets/packets.rs
```diff
@@ -708,6 +708,240 @@ impl CfgMsgAllPortsBuilder {
...
+    pdop: u16,
+
+    /// Time DOP Mask to use
+    #[ubx(map_type = f32, scale = 0.1, into = ScaleBack::<f32>(10.).as_u16)]
```
lkolbly: If we know the map_type and the scale, can we not automatically generate the into function, rather than having to separately specify the function?

ublox/src/ubx_packets/packets.rs
Dushistov: > We should be consistent about using f64 (or f32, but my preferences lean toward f64).

I supposed we pretty consistent about this. Rule is simple - minimal type that allow data-without-loss conversation. It is impossible convert `u32/i32` to `f32` without data loss, so I used `f64`, it is possible to convert `u16` to `f32` without data loss, so I used `f32` and so on.

I suppose this is consistent and use the possible minimal of computer resources.

Dushistov:
In fact, during implementation of this I think in ideal world it should not be mapping
to `f32` / `f64`. It should be something like this:

```rust
struct ScaledNumberMul<T, coef: uX1>(T);
struct ScaledNumberDiv<T, coef: uX2>(T);
```
where `ScaledNumberMul -> fX3 = self.0 * fX3::from(coef)`, and    `ScaledNumberDiv -> fX4 = self.0 / fX4::from(coef)`.

then it would be possible to skip usage of float-point arithmetic for some cases.
For other cases it would still possible to use via `impl From<ScaledNumberMul> for f32/f64`.
But to implement it in zero-cost way, there is need for generic const / rust-lang/rust#44580 in stable.

ublox/src/ubx_packets/packets.rs
lkolbly: At the point where we're converting things into floating point numbers, I would prefer that the user doesn't have to care at all about the internal layout of the packet. If we switch back and forth between f32 and f64, we're forcing the user to care about whether the underlying field was represented as a u16 or as a u32. The user just wants to take their value and do their processing.

You raise a good point that 32-bit ints can only be fully represented as f64, so let's stick with f64.

lkolbly:
Yeah, someday ideally we'll have good support for platforms without FPUs, through some sort of fixed-point type. Googling around there don't seem to be any base-10-fixed-point-type crates, but as you say we could hack a custom one.

ublox/src/ubx_packets/packets.rs
Dushistov: Done

ublox/src/ubx_packets/packets.rs
Dushistov: > If we switch back and forth between f32 and f64, we're forcing the user to care about whether > the underlying field was represented as a u16 or as a u32
> The user just wants to take their value and do their processing.

Didn't get this point. 

1. The UBX protocol is rather consistent, so with consistent rule "minimal float type without data loss" our representation would be also rather consistent. For example pdop.
`NavSolution::pdop` is `f32` and `CfgNav5::pdop` is `f32`. So user should not care about
converting type, he/she will get the same type for the same value. It would be problem converting latitude to GDOP, but I suppose this is good thing that helps user catch errors.

2. Obviously user should care about underlying types. In "sending" case. If he/she put to big value, it would be silently truncated, and user will have bad time finding out what is going on,
 so better to show user via types that this field do not accept so "huge" value.

3. Sometimes CPU can handle `f32` in more fast way then `f64`, why force generation of slower code?

4. We don't use `i64` / `u64` everywhere for any integer type, why use `f64` everywhere for every float-point type?

lkolbly: > In "sending" case. If he/she put too big value, it would be silently truncated, and user will have bad time finding out what is going on, so better to show user via types that this field do not accept so "huge" value.

This doesn't really apply to f32/f64: f32 and f64 both can represent values which cannot be exactly represented in the corresponding fixed-point types, and vice-versa.

That's also why it doesn't really make sense to use i64/u64 everywhere, since in those cases the types are being exactly preserved. (setting aside the fact that it makes the codegen easier)

The user's application will almost certainly be either all-f32 or all-f64 (in my own experience, I find it is uncommon to mix and match, and when the codebase isn't consistent I find it painful). We can't predict which type the user will want to use, but we can at least present a uniform interface to them so that they can either always cast or never cast. f64 is capable of exactly representing all values, and the performance impact is negligible*, so f64 seems like the path to take.

\*  Even if we're transferring a constant 1Mb/s of u16s which we interpret as doubles, that's still a mere 62K doubles/s which we need to convert - and we're already doing a floating-point multiplication (or division!) on every value.

Dushistov: > This doesn't really apply to f32/f64

why, I doubt that user of library would use 2^100, but
he/she can use something that do not fit into mantissa of `f32`,
but fit into `f64`, if he/she have no idea what exactly what possible values 
can be used. Yes `f32` and `f64` don't exact restriction, but it at least some restriction.

> The user's application will almost certainly be either all-f32 or all-f64
> in my own experience,

You made logic mistake here. It is impossible to come from particular case to general case, if your experience don't cover all existing programs.
Last two examples from my experience:

1. Spatial object search, at first search with [r-tree](https://en.wikipedia.org/wiki/R-tree) and then loop over all variants. R-tree was `f32` based, because of memory consumption, CPU usage and so things and this part of search can be "rough", and then calculation of geodesic distance  done with `f64`, because of this part should be precise. 

2. Drawing and calculation. Core of program was based on `f64` calculations, while for drawing was used `f32`, basically it can be `i32`, but `f32` was used for perfect line and font aliasing. 

> but we can at least present a uniform interface to them so that they can either always cast or never cast

This is inconsistent with your words:

> Yeah, someday ideally we'll have good support for platforms without FPUs, through some sort of fixed-point type.

So let's suppose Rust will allow this:
```Rust
struct ScaledNumber<T, Mul: u16>(T);
impl From<ScaledNumber<T, Mul: u16>> for f64 {
   fn from(x: ScaledNumber<T, Mul: u16>) -> f64 {
       f64::from(x.0) * f64::from(Mul)
   }
}
```

and user of our library will right:
```Rust
   let x: f64 = packet_ref.pdop().into();
```

but this exactly the same that he/she should write now:
```rust
// fn pdop(&self) -> f32;
let x: f64 = packet_ref.pdop().into();
```

So from user point of view the interface will be exactly the same.
But the second one is bad, while the first one is good?

lkolbly: > R-tree was f32 based, because of memory consumption, CPU usage and so things and this part of search can be "rough", and then calculation of geodesic distance done with f64, because of this part should be precise.

Even in this situation, though, the part that interacts with the GPS I imagine will be all one way or all the other (I'm imagining an application that locates a ublox devices inside a r-tree. Or something). You're right that a single crate could use different types for different reasons, but I'd prefer not to force that onto them.

> But the second one is bad, while the first one is good?

The first one is good, because the user always calls `into()` to get to their preferred type.

The second one is bad, because the user code will look like:
```
let x: f64 = packet_ref.foo().into();
let y: f64 = packet_ref.bar();
let z: f64 = packet_ref.baz().into();
```
When I write code, I don't call .into() by default. If the user then needs to get field "barz", it's 50/50 whether they need to convert, and it's not clear except by either looking at the docs or just trying it and seeing if it compiles.

Dushistov:
@lkolbly 

I think with macro_rules or/and procedural macros it is possible to generate something like:

```rust
struct ScaledNumBy100Mul(u32);

impl ScaledNumBy100Mul {
   const SCALE: u8 = 100;
}

impl From<ScaledNumBy100Mul> for f32/f64
```

It would be not as great as usage of const generics, but should be quite usable
and should resolve our disagreement about f32/f64.
But I had no time for implementation.

lkolbly:
@Dushistov Yeah, I've been busy too, I pulled this in since I didn't feel it was important enough to block forever. Someday before 1.0 we'll think of the right unified way to do this.

Speaking of which, let me know if/when you would like a release, I have no pending features that I want for a 0.2.0, I can arrange to double-check the code on my device and make a release.

PR #14, opened 2021-06-25
Add 'alloc' feature

jbeaurivage:
# Add the 'alloc' Cargo feature

The aim of this PR is to enable true allocator-free compilation for `#![no_std]` targets.

Now that arrays can be directly used as buffers for a `Parser`, it would be nice if the requirement to use a global allocator would be removed entirely for `#![no_std]` targets. This PR adds an `alloc` Cargo feature, enabling the avoiding relying on the `alloc` crate entirely.

### DIFF
--- a/ublox/Cargo.toml
+++ b/ublox/Cargo.toml
@@ -11,6 +11,7 @@ readme = "../README.md"
 [features]
 default = ["std"]
 std = []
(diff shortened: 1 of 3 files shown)


lkolbly (approved):
This looks good, thanks!

PR #15, opened 2021-08-30
Add various UBX messages

aholtzma-am:
This change adds the following packet definitions:

UBX-MON-HW
UBX-CFG-ANT
UBX-CFG-NAVX5
UBX-CFG-INF
UBX-INF
UBX-MGA-ACK

### DIFF
--- a/ublox/src/ubx_packets/packets.rs
+++ b/ublox/src/ubx_packets/packets.rs
@@ -513,6 +513,14 @@ struct CfgRst {
     reserved1: u8,
 }
 
+/// Reset Receiver / Clear Backup Data Structures
+#[ubx_packet_recv_send]
+#[ubx(class = 6, id = 0x13, fixed_payload_len = 4)]
+struct CfgAnt {
+    flags: u16,
+    pins: u16,
+}
+
 #[ubx_extend_bitflags]
 #[ubx(into_raw, rest_reserved)]
 bitflags! {
@@ -1115,6 +1123,7 @@ define_recv_packets!(
         CfgPrtSpi,
         CfgPrtUart,
         CfgNav5,
+        CfgAnt,
         MonVer,
         MonHw
     }
--- a/ublox/src/ubx_packets/packets.rs
+++ b/ublox/src/ubx_packets/packets.rs
@@ -433,11 +433,11 @@ bitflags! {
     /// `CfgInfMask` parameters bitmask
     #[derive(Default)]
     pub struct CfgInfMask: u8 {
-        const ERROR= 0x1;
+        const ERROR = 0x1;
         const WARNING = 0x2;
-        const NOTICE  = 0x4;
-        const TEST  = 0x8;
-        const DEBUG  = 0x10;
+        const NOTICE = 0x4;
+        const DEBUG = 0x08;
+        const TEST  = 0x10;
     }
 }
 
--- a/ublox/src/ubx_packets/packets.rs
+++ b/ublox/src/ubx_packets/packets.rs
@@ -402,6 +402,45 @@ enum NavStatusFlags2 {
     Inactive = 3,
 }
 
+/// Information message conifg
+#[ubx_packet_recv_send]
+#[ubx(
+    class = 0x06,
+    id = 0x2,
+    fixed_payload_len = 10,
+    flags = "default_for_builder"
+)]
+struct CfgInf{
+    protocol_id: u8,
+    reserved: [u8; 3],
+    #[ubx(map_type = CfgInfMask)]
+    inf_msg_mask_0: u8,
+    #[ubx(map_type = CfgInfMask)]
+    inf_msg_mask_1: u8,
+    #[ubx(map_type = CfgInfMask)]
+    inf_msg_mask_2: u8,
+    #[ubx(map_type = CfgInfMask)]
+    inf_msg_mask_3: u8,
+    #[ubx(map_type = CfgInfMask)]
+    inf_msg_mask_4: u8,
+    #[ubx(map_type = CfgInfMask)]
+    inf_msg_mask_5: u8,
+}
+
+#[ubx_extend_bitflags]
+#[ubx(from, into_raw, rest_reserved)]
+bitflags! {
+    /// `CfgInfMask` parameters bitmask
+    #[derive(Default)]
+    pub struct CfgInfMask: u8 {
+        const ERROR= 0x1;
+        const WARNING = 0x2;
+        const NOTICE  = 0x4;
+        const TEST  = 0x8;
+        const DEBUG  = 0x10;
+    }
+}
+
 #[ubx_packet_send]
 #[ubx(
     class = 0x0B,
--- a/ublox/src/ubx_packets/packets.rs
+++ b/ublox/src/ubx_packets/packets.rs
@@ -525,10 +525,14 @@ struct InfDebug{
 
 mod inf {
     pub(crate) fn convert_to_str(bytes: &[u8]) -> &str {
-        core::str::from_utf8(bytes).unwrap()
+        match core::str::from_utf8(bytes) {
+            Ok(msg) => msg,
+            Err(_) => "** Invalid UTF8 string in UBX-INF packet **",
+        }
     }
 
-    pub(crate) fn is_valid(bytes: &[u8]) -> bool {
+    pub(crate) fn is_valid(_bytes: &[u8]) -> bool {
+        // Validity is checked in convert_to_str
         true
     }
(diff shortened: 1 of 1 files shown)


lkolbly:
Hi, thanks for the PR! I will take a look hopefully Wednesday evening (9/1).

ublox/src/ubx_packets/packets.rs
```diff
@@ -402,6 +402,137 @@ enum NavStatusFlags2 {
...
+#[ubx(from, into_raw, rest_reserved)]
+bitflags! {
+    /// `CfgInfMask` parameters bitmask
+    #[derive(Default)]
+    pub struct CfgInfMask: u8 {
+        const ERROR= 0x1;
+        const WARNING = 0x2;
+        const NOTICE  = 0x4;
+        const TEST  = 0x8;
+        const DEBUG  = 0x10;
```
lkolbly, line 440: Looking at the u-blox 7 protocol spec, it looks like these two might be swapped?
image

ublox/src/ubx_packets/packets.rs
```diff
@@ -402,6 +402,137 @@ enum NavStatusFlags2 {
...
+    inf_msg_mask_4: u8,
+    #[ubx(map_type = CfgInfMask)]
+    inf_msg_mask_5: u8,
+}
+
+#[ubx_extend_bitflags]
+#[ubx(from, into_raw, rest_reserved)]
+bitflags! {
+    /// `CfgInfMask` parameters bitmask
+    #[derive(Default)]
```
lkolbly, line 434: What's the default of a bitflags struct? zero? (and, if so, is zero a valid value for one of those fields?)

ublox/src/ubx_packets/packets.rs
```diff
@@ -402,6 +402,137 @@ enum NavStatusFlags2 {
...
+
+
+mod inf {
+    pub(crate) fn convert_to_str(bytes: &[u8]) -> &str {
+        core::str::from_utf8(bytes).unwrap()
+    }
+
+    pub(crate) fn is_valid(bytes: &[u8]) -> bool {
+        true
+    }
```
lkolbly, line 533: Panicking isn't an ideal approach (although I can't claim that this crate is currently panic-free), since it limits the user's options in the case of corrupt data. Would it be possible to fill in `is_valid` like so:
```
core::str::from_utf8(bytes).is_ok()
```
?

If performance is an issue, we could make `convert_to_str` return an option:
```
core::str::from_utf8(bytes).ok()
```
and set `map_type` to `Option<&str>`. (and also remove the `may_fail`, I think)

ublox/src/ubx_packets/packets.rs
```diff
@@ -1020,6 +1159,163 @@ impl<T: FloatCore + FromPrimitive + ToPrimitive> ScaleBack<T> {
...
+    reserved2: u8,
+
+    /// initial fix must be 3D
+    ini_fix_3d: u8,
+
+    /// Reserved
+    reserved3: [u8; 2],
+
+    /// issue acknowledgements for assistance message input
+    ack_aiding: u8,
```
lkolbly, line 1203: Just for my own understanding, if you don't mind my asking, do you know what version of the protocol you're using? I haven't yet decided what the best way to handle version discrepancies is, but until then I'm trying to get a sense of what versions people are using.

ublox/src/ubx_packets/packets.rs
```diff
@@ -1020,6 +1159,163 @@ impl<T: FloatCore + FromPrimitive + ToPrimitive> ScaleBack<T> {
...
+    /// GPS week rollover number
+    wkn_rollover: u16,
+
+    /// Permanently attenuated signal compensation
+    sig_atten_comp_mode: u8,
+
+    /// Reserved
+    reserved4: u8,
+    reserved5: [u8; 2],
+    reserved6: [u8; 2],
```
lkolbly, line 1214: Why do you separate this into three different fields? (rather than `[u8; 5]`)

ublox/src/ubx_packets/packets.rs
```diff
@@ -1020,6 +1159,163 @@ impl<T: FloatCore + FromPrimitive + ToPrimitive> ScaleBack<T> {
...
+        const USE_ADR = 0x40;
+        ///  apply signal attenuation compensation feature settings
+        const USE_SIG_ATTEN_COMP = 0x80;
+    }
+}
+
+/// Hardware status
+#[ubx_packet_recv]
+#[ubx(class = 0x0a, id = 0x09, fixed_payload_len = 60)]
+struct MonHw{
```
lkolbly, line 1275: (nitpick)
```suggestion
struct MonHw {
```

lkolbly (changes requested):

ublox/src/ubx_packets/packets.rs
aholtzma-am: Good catch. Fixed.

ublox/src/ubx_packets/packets.rs
aholtzma-am: It is indeed zero, and zero is a valid mask.

ublox/src/ubx_packets/packets.rs
aholtzma-am: We use a number of different models of GPS with protocol versions 15, 17, and 22. Handling multiple versions of the same message is going to be messy, I'm not sure what the best approach is. So far we haven't had to deal with it.

ublox/src/ubx_packets/packets.rs
aholtzma-am: I've fixed this by substituting an error string if UTF8 conversion fails.

ublox/src/ubx_packets/packets.rs
aholtzma-am: So this is how it is defined in the specification. This is likely just an undocumented field. If someone wanted to RE these fields, at least they are pulled out correctly here.

ublox/src/ubx_packets/packets.rs
```diff
@@ -402,6 +402,141 @@ enum NavStatusFlags2 {
...
+        from = inf::convert_to_str,
+        get_as_ref)]
+    message: [u8; 0]
+}
+
+#[ubx_packet_recv]
+#[ubx(
+    class = 0x4,
+    id = 0x2,
+    max_payload_len = 1240,
```
aholtzma-am, line 464: One loose end here is that there is no max payload length for these INF messages. How to express this? The code generator wants either fixed or max payload length.

ublox/src/ubx_packets/packets.rs
aholtzma-am: Fixed.

ublox/src/ubx_packets/packets.rs
lkolbly: This is certainly suboptimal. I think this is fine for now though, eventually I will change the code generator to handle this better (and I'll update this when I get around to that).

ublox/src/ubx_packets/packets.rs
lkolbly: Playing around with it a bit locally, I'm a little worried about the fact that the user has no type information to indicate whether decoded succeeded (after all, in principle, the device could send a log message "** Invalid UTF8 string in UBX-INF packet **")

So, I'm thinking we can do this:
```
#[ubx_packet_recv]
#[ubx(
    class = 0x4,
    id = 0x4,
    max_payload_len = 1240,
    flags = "default_for_builder"
)]
struct InfDebug{
    #[ubx(map_type = Option<&str>,
        from = inf::convert_to_str,
        get_as_ref)]
    message: [u8; 0]
}
```
and:
```
mod inf {
    pub(crate) fn convert_to_str(bytes: &[u8]) -> Option<&str> {
        match core::str::from_utf8(bytes) {
            Ok(msg) => Some(msg),
            Err(_) => None,
        }
    }
}
```
and then, for the user, we get to keep the same API (they can get their message with `packet.message().unwrap()`)

lkolbly:
Hey! Sorry, I meant to follow up on this a while ago but I got distracted and then never came back :P

I think the only outstanding thought I had on this was the error handling in `convert_to_str`, unless it doesn't work for you I think I will pull in this PR as-is and then add a commit to map it to an Option or a Result or something.

aholtzma-am:
I was off working on other things and forgot about it too. I will make the
requested changes tonight.

lkolbly:
This looks good, thanks! I'll put this into a `0.3.1` release along with the `NavSat` changes.

PR #21, opened 2022-05-12
`str` should not include NULL byte.

reitermarkus:
Also, why is `convert_to_str_unchecked` called this when in fact it is very much checked?


lkolbly (approved):
Thanks for the PR! Sorry for the delay.

> Also, why is convert_to_str_unchecked called this when in fact it is very much checked?

Because it might panic, so the `_unchecked` is intended to emphasize that `is_cstr_valid` should be called first.

PR #23, opened 2022-05-31
Add `CfgValSet` and `RxmRawx` packets.

reitermarkus:
- Add `CFG_VAL_SET` command and some of the values that can be set using it.
- Add `RXM_RAWX` packet.

### DIFF
--- a/ublox/src/ubx_packets/cfg_val.rs
+++ b/ublox/src/ubx_packets/cfg_val.rs
@@ -1,4 +1,4 @@
-use super::{CfgInfMask, DataBits, Parity, StopBits};
+use super::{AlignmentToReferenceTime, CfgInfMask, DataBits, Parity, StopBits};
 
 pub struct KeyId(u32);
 
@@ -52,6 +52,12 @@ macro_rules! from_cfg_v_bytes {
             _ => unreachable!(),
         }
     };
+    ($buf:expr, u8) => {
+        $buf[0]
+    };
+    ($buf:expr, u16) => {
+        u16::from_le_bytes([$buf[0], $buf[1]])
+    };
     ($buf:expr, u32) => {
         u32::from_le_bytes([$buf[0], $buf[1], $buf[2], $buf[3]])
     };
@@ -82,6 +88,16 @@ macro_rules! from_cfg_v_bytes {
             _ => unreachable!(),
         }
     };
+    ($buf:expr, AlignmentToReferenceTime) => {
+        match $buf[0] {
+            0 => AlignmentToReferenceTime::Utc,
+            1 => AlignmentToReferenceTime::Gps,
+            2 => AlignmentToReferenceTime::Glo,
+            3 => AlignmentToReferenceTime::Bds,
+            4 => AlignmentToReferenceTime::Gal,
+            _ => unreachable!(),
+        }
+    };
 }
 
 macro_rules! into_cfg_kv_bytes {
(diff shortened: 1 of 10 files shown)


ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -0,0 +1,438 @@
...
+      into_cfg_kv_bytes!(@inner [
+          $this.0 as u8
+      ])
+    };
+}
+
+macro_rules! cfg_val {
+  ($($cfg_item:ident, $cfg_key_id:expr, $cfg_value_type:ident,)*) => {
+    #[derive(Debug, Clone, Copy)]
+    pub enum CfgVal {
```
reitermarkus, line 201:
```suggestion
    #[non_exhaustive]
    pub enum CfgVal {
```

lkolbly:
Hey! Sorry for the delay. This is a good PR, I just haven't had enough time to sit down and look through it. I'm not going to be able to look at it for the next couple weeks, so I appreciate your patience in the meantime. Don't worry, it is on my list. :-)

reitermarkus:
@lkolbly, ping on this. Hopefully you have some time to review this now. 

If not, maybe consider adding another maintainer since there are also a bunch of other open PRs?

ublox/src/ubx_packets/packets.rs
```diff
@@ -1825,6 +1899,321 @@ pub enum MsgAckInfoCode {
     RejectedUnknownType = 6,
 }
 
+/// Time pulse time data
+#[ubx_packet_recv]
+#[ubx(class = 0x0d, id = 0x01, fixed_payload_len = 16)]
+struct TimTp {
+  /// Time pulse time of week according to time base
```
lkolbly, line 1906: Can you run `cargo fmt` on this? (I suspect the indentation will change)

ublox_derive/src/output.rs
```diff
@@ -227,36 +235,56 @@ pub fn generate_send_code_for_packet(pack_descr: &PackDesc) -> TokenStream {
...
         };
-        if let Some(into_fn) = f.map.map_type.as_ref().map(|x| &x.into_fn) {
-            pack_fields.push(quote! {
-                let bytes = #into_fn(self.#name).to_le_bytes()
-            });
-        } else if !f.is_field_raw_ty_byte_array() {
-            pack_fields.push(quote! {
-                let bytes = self.#name.to_le_bytes()
+
+        if size_bytes == 0 {
```
lkolbly, line 242: I think we could also flip this if statement, and have:
```
if let Some(size_bytes) = f.size_bytes {
    // everything currently in the else condition and after
    if let Some(into_fn) = .... {
    }
    ...
    off += size_bytes;
} else {
    extend_fields.push(...);
}
```
and then we don't have to map from None to 0, and there isn't that continue splitting up the code.

But, bigger question: because of the `off += size_bytes` that we're skipping (because we don't know the actual size_bytes), will this code break if this field is not the last one?

If so, should we just assert that this is the last field, and break?

lkolbly:
> @lkolbly, ping on this. Hopefully you have some time to review this now.
> 
> If not, maybe consider adding another maintainer since there are also a bunch of other open PRs?

Hey! Sorry, yes. I am still alive. Obviously, a couple weeks turned into a couple months (as they always do...).

I left some comments. Let me know if you get time to look at them, if not it's perfectly fine I can merge this as-is and handle it on my end.

reitermarkus:
@lkolbly, I flipped the `if` and added the assertion that the iterator is the last field.

PR #30, opened 2022-07-06
add feature for serdes

gwbres:
Hello @lkolbly,
I'm building an application that involves an `ublox` and I need the user to control and parametrize it.
Having the `serdes` capacity would really help. With this commit I introduce a new `use-serde` build feature and some serdes derivation. I am mostly interested in `GpsFix`, `Odo` options and `Ant` options

### DIFF
--- a/ublox/Cargo.toml
+++ b/ublox/Cargo.toml
@@ -12,7 +12,6 @@ readme = "../README.md"
 default = ["std"]
 std = []
 alloc = []
-use-serde = ["serde"]
 
 [dependencies]
 chrono = { version = "0.4.19", default-features = false, features = [] }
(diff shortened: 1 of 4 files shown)


ublox/Cargo.toml
```diff
@@ -12,12 +12,14 @@ readme = "../README.md"
 default = ["std"]
 std = []
 alloc = []
+use-serde = ["serde"]
```
reitermarkus, line 15: You can just leave this out, I think there will automatically be a `serde` feature for the optional dependency.

gwbres: Hello @reitermarkus,
I confirm, this works

ublox/src/lib.rs
```diff
@@ -69,6 +69,10 @@
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
+#[cfg(feature = "use-serde")]
```
reitermarkus, line 72: The feature should be called `serde` instead of `use-serde`.

gwbres: woops, I forgot this, implied by your 1st review,
good catch

PR #40, opened 2022-09-14
Serde for Iterators, Added new messages

adrian-kong:
This PR adds some of missing messages (before recent merges) and code-gen for serde serialize implementations, helpful for serializing iterators.

List of new messages:
- ESF_MEAS,
- RXM_RAWX
- RXM_SFRBX
- ESF_RAW
- ESF_INS
- HNR_PVT
- NAV_ATT
- NAV_CLOCK
- NAV_VELECEF
- MGA_GPA_EPH

Serialization outputs includes the class id and message id*


.github/workflows/rust.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - name: Install MSRV
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.57.0
+        toolchain: 1.63.0
```
lkolbly, line 24: Hey! Sorry for the delayed review.

Do we need 1.63 for this change, or can we get by with 1.57? I'm trying to keep the MSRV around a year old or so.

adrian-kong, line 24:
```suggestion
        toolchain: 1.57.0
```

adrian-kong: I don't think so, can try reverting this. Besides that, is everything all good?

ublox_cli/src/main.rs
```diff
@@ -1,196 +1,20 @@
...
-                if let PacketRef::AckAck(ack) = packet {
-                    if ack.class() == T::CLASS && ack.msg_id() == T::ID {
-                        found_packet = true;
-                    }
-                }
-            })?;
-        }
-        Ok(())
-    }
+use std::io;
```
lkolbly, line 1: Would it make sense to move this to a different example crate, or add it as another option alongside the previous code?

.github/workflows/rust.yml
lkolbly: I just commented about the ublox_cli changes, but then yeah everything looks good. (it looks like there might be some merge conflicts, I haven't gotten a chance to look at them yet)

ublox_cli/src/main.rs
adrian-kong: Yeah was just testing for our use case, this has been reverted though. Passes CI as well with reverted MSRV

ublox/Cargo.toml
```diff
@@ -9,7 +9,7 @@ repository = "https://github.com/lkolbly/ublox"
 readme = "../README.md"
 
 [features]
-default = ["std"]
+default = ["std", "serde"]
```
silverjam, line 12: Should we disable serde by default?

.github/workflows/rust.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install libudev
-      run: sudo apt-get install -y libudev-dev
+      run: sudo apt-get update && sudo apt-get install -y libudev-dev
```
silverjam, line 20: We need a build variant that ensures that the library can build with `serde` enabled, correct?

silverjam (commented):
@lkolbly Can you please enable CI checks for this PR?

lkolbly:
@silverjam I think CI is currently enabled for this PR?

(also, I should be able to merge this this weekend, hopefully)

lkolbly:
There are some merge conflicts though, maybe that's blocking CI.

ublox/Cargo.toml
lkolbly: I think serde by default should be fine.

.github/workflows/rust.yml
lkolbly: We also need a matrix that tests we can build no_std and no alloc, for now I think we can survive with just testing the default case and we'll add that later.

lkolbly:
@silverjam Sorry for the delay, yeah apparently it wasn't enabled, not sure what was up with that. Anyway I enabled CI and it looks like we just need to `use serde::{Serialize, Deserialize}`

silverjam:
@lkolbly Ok, builds and tests are passing locally for me -- looks like you'll need to approve the CI one more time

silverjam:
Thanks for getting this in @lkolbly!

PR #41, opened 2022-09-16
packets.rs: introduce CfgItfm frame

gwbres:
Hello @lkolbly, @reitermarkus,

The CFG_ITFM frame is an important frame to our application, it controls the Interference (jamming) detection module.

I would like to control the `Default` value of this frame myself, as some part of the config words (Algorithm & config2) must have very specific (and undocumented values). This is how I did it

### DIFF
--- a/ublox/src/ubx_packets/packets.rs
+++ b/ublox/src/ubx_packets/packets.rs
@@ -741,6 +741,252 @@ impl Default for OdoProfile {
     }
 }
 
+/// Configure Jamming interference monitoring 
+#[ubx_packet_recv_send]
+#[ubx(class = 0x06, id = 0x39, fixed_payload_len = 8)]
+struct CfgItfm {
+    /// Interference config Word
+    #[ubx(map_type = CfgItfmConfig)]
+    config: u32,
+    /// Extra settings
+    #[ubx(map_type = CfgItfmConfig2)]
+    config2: u32,
+}
+
+#[derive(Debug, Copy, Clone)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct CfgItfmConfig {
+    /// enable interference detection
+    enable: bool,
+    /// Broadband jamming detection threshold (dB)
+    bb_threshold: CfgItfmBbThreshold,
+    /// CW jamming detection threshold (dB)
+    cw_threshold: CfgItfmCwThreshold,
+    /// Reserved algorithm settings
+    /// should be set to 0x16B156 default value
+    /// for correct settings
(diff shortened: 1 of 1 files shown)


maximeborges:
I think doing something similar to `CfgPrtUart` with `UartMode` would fit better in this code and be more Rustic.

Something along those lines:
```rust
#[ubx_packet_recv_send]
#[ubx(
    class = 0x06, 
    id = 0x39, 
    fixed_payload_len = 8,
    flags = "default_for_builder"
)]
struct CfgItfm {
    /// Interference config Word
    #[ubx(map_type = CfgItfmConfig)]
    config: u32,

    /// Extra settings
    #[ubx(map_type = CfgItfmConfig2)]
    config2: u32,
}

#[derive(Debug, Copy, Clone)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
pub struct CfgItfmConfig {
    /// Broadband jamming detection threshold (dB)
    bb_threshold: BbThreshold,
    /// CW jamming detection threshold (dB)
    cw_threshold: CwThreshold,
    /// Reserved algorithm settings - should be set to 0x16B156 in hex for correct settin
    algorithm_bits: AlgorithmBits,
    /// enable interference detection
    enable: bool,
}

#[derive(Debug, Clone, Copy)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
pub struct BbThreshold(u32);
impl BbThreshold {
    const POSITION: u32 = 0;
    const LENGTH: u32 = 4;
    const MASK: u32 = (1<<Self::LENGTH)-1;

    const fn into_raw(self) -> u32 {
        (self.0 & Self::MASK) << Self::POSITION
    }
}
impl From<u32> for BbThreshold {
    fn from(thres: u32) -> Self {
        Self(thres)
    }
}
```

And so on.

gwbres:
> I think doing something similar to `CfgPrtUart` with `UartMode` would fit better in this code and be more Rustic.

Yes that would work, I will update this PR accordingly

gwbres:
Hello,

@maximeborges for information, the code did not compile as is, because every single structures involved in the `ubx_recv/send` generators require an `into_raw()` method. 

I also wanted to keep the `boolean` native types for enable bits, because I think it's easier to use, at the user level.
Finally, I also wanted to use an `enum` not a struct wrapper for the `AntennaSettings`, to emphasize that only those values are correct. I use the `[repr(u8)]` derivation, and cast it to `u32` in the needed `into_raw()` methods

lkolbly (approved):
This looks good to me, thank you!

PR #42, opened 2022-09-22
Add new messages in cfg_val and add support for NavHpPosLlh

maximeborges:
This PR adds support for more messages for CFG-VAL-SET, and also add a handler for UBX-NAV-HPPOSLLH for high precision positioning.

I also added a helper for `CfgValSetBuilder` since it wasn't very obvious how to use it before, and some examples in the README. The `cfg_val!` macro now also supports documentation comment.

I'm not sure it was a good idea to add both cfg_val and NavHpPosLlh modifications in a single PR, I can split it up if it makes more sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,31 @@ let packet: [u8; 28] = CfgPrtUartBuilder {
    reserved5: 0,
 }.into_packet_bytes();
 ```
+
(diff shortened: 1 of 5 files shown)


lkolbly:
CI failure looks like an issue with CI, you can ignore that.

lkolbly (approved):
This looks good, thank you!

Issue #43, opened 2022-09-25
Proper Versioning Support


lkolbly:
This project has so far been really bad about handling different versions of the protocol. For examples, see #33 and #24. The goal of this project is to support "all" versions of the protocol, but so far the API pretends that all protocols are the same.

I'm kind of leaning towards defining a module per version. For example, an import might look like:
```
use ublox::m8::NavRelPosNed;
```

This has the advantage that there is a clear delineation of protocols, and what packets are supported in which protocol. But at the same time, it means that extra work must happen to take advantage of similarities between protocols. If the user just wants to parse some simple packet that is common across many versions, and they want their binary to be somewhat device-agnostic, then they must separately parse each version:
```
if talking_to_m8 {
    // Parse a ublox::m8::NavPosLlh
} else if talking_to_f9p {
    // Parse a ublox::f9p::NavPosLlh
}
```

Additionally, although there is a clear delineation across broad-stroke "m8", "m7", "f9p" labels, there isn't much ability to focus on sub-product version changes. For example, UBX-CFG-TMODE3 on M8 is only supported in protocol versions 20.x, even though "m8" is protocol versions 15 through 23. So the packet would appear to be supported, and the parser would be generated, for all "m8" products. I think that might be fine though.

Anyway. I'm creating this issue so that there's one place to discuss this issue as a whole. My intent is that this is the next minor-bumping release (so, 0.5.0).

gwbres:
> I'm kind of leaning towards defining a module per version. For example, an import might look like:
> 
> ```
> use ublox::m8::NavRelPosNed;
> ```

that approach looks good, at the expense of a little more complexity and code volume: we need some kind of wrapper to determine whether this frame is revision dependent or is shared across revisions.  I assume in that scenario it would be interesting to turn revisions into compilation features ?

andrei-ng:
Added version support as of PR https://github.com/ublox-rs/ublox/pull/87 . Closing this.

PR #44, opened 2022-09-26
Add PacketOwned variants

maximeborges:
This PR allows to copy parsed packets into data-owned packets that can be moved to other threads.

I'm not really happy about duplicating all the packets and having big matching tables here, but I'm not really sure how else this could be implemented. Maybe refactoring so you have `PacketOwned` as a base data structure that you can borrow from to get `PacketRef`.

Also unintentionally run `cargo fmt` so it cleaned up some other code, which is not that bad in the end (but I can split that into two commits at least if it helps for the review).

### DIFF
--- a/ublox/src/ubx_packets.rs
+++ b/ublox/src/ubx_packets.rs
@@ -83,9 +83,17 @@ pub trait UbxPacketCreator {
 }
 
 /// Packet not supported yet by this crate
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug)]
 pub struct UbxUnknownPacketRef<'a> {
     pub payload: &'a [u8],
+    pub class: u8,
+    pub msg_id: u8,
+}
+
+/// Packet not supported yet by this crate
+#[derive(Clone, Debug)]
+pub struct UbxUnknownPacketOwned {
+    pub payload: Vec<u8>,
     pub class: u8,
     pub msg_id: u8,
 }
--- a/ublox_derive/src/output.rs
+++ b/ublox_derive/src/output.rs
@@ -686,7 +686,9 @@ pub fn generate_code_for_parse(recv_packs: &RecvPackets) -> TokenStream {
         });
         matches_owned.push(quote! {
             (#name::CLASS, #name::ID) if <#owned_name>::validate(payload).is_ok()  => {
-                Ok(#union_enum_name_owned::#name(#owned_name([0; #owned_name::PACKET_SIZE])))
+                let mut bytes = [0u8; #owned_name::PACKET_SIZE];
+                bytes.clone_from_slice(payload);
+                Ok(#union_enum_name_owned::#name(#owned_name(bytes)))
             }
         });
         matches_ref_to_owned.push(quote! {
--- a/ublox_derive/src/output.rs
+++ b/ublox_derive/src/output.rs
@@ -8,7 +8,12 @@ use quote::{format_ident, quote};
 use std::{collections::HashSet, convert::TryFrom};
 use syn::{parse_quote, Ident, Type};
 
-fn generate_debug_impl(pack_name: &str, ref_name: &Ident, pack_descr: &PackDesc) -> TokenStream {
+fn generate_debug_impl(
+    pack_name: &str,
+    ref_name: &Ident,
+    owned_name: &Ident,
+    pack_descr: &PackDesc,
+) -> TokenStream {
     let mut fields = vec![];
     for field in pack_descr.fields.iter() {
         let field_name = &field.name;
@@ -26,12 +31,25 @@ fn generate_debug_impl(pack_name: &str, ref_name: &Ident, pack_descr: &PackDesc)
                     .finish()
             }
         }
+
+        impl core::fmt::Debug for #owned_name {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                f.debug_struct(#pack_name)
+                    #(#fields)*
+                    .finish()
+            }
+        }
     }
 }
 
 pub fn generate_recv_code_for_packet(pack_descr: &PackDesc) -> TokenStream {
     let pack_name = &pack_descr.name;
     let ref_name = format_ident!("{}Ref", pack_descr.name);
+    let owned_name = format_ident!("{}Owned", pack_descr.name);
+    let packet_size = match pack_descr.header.payload_len {
+        PayloadLen::Fixed(value) => value,
+        PayloadLen::Max(value) => value,
+    } as usize;
 
     let mut getters = Vec::with_capacity(pack_descr.fields.len());
     let mut field_validators = Vec::new();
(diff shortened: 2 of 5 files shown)


gwbres:
@maximeborges very good contribution :+1:

ublox/src/ubx_packets.rs
```diff
@@ -83,13 +83,21 @@ pub trait UbxPacketCreator {
...
 pub struct UbxUnknownPacketRef<'a> {
     pub payload: &'a [u8],
     pub class: u8,
     pub msg_id: u8,
 }
 
+/// Packet not supported yet by this crate
+#[derive(Clone, Debug)]
+pub struct UbxUnknownPacketOwned {
+    pub payload: Vec<u8>,
```
lkolbly, line 96: Is it possible to put this behind the `alloc` flag? (for clients who haven't defined an allocator)

ublox_derive/src/output.rs
```diff
@@ -635,30 +716,69 @@ pub fn generate_code_for_parse(recv_packs: &RecvPackets) -> TokenStream {
...
+                })),
+            }
+        }
+
+        impl<'a> From<&#union_enum_name_ref<'a>> for #union_enum_name_owned {
+            fn from(packet: &#union_enum_name_ref<'a>) -> Self {
+                match packet {
+                    #(#matches_ref_to_owned)*
+                    #union_enum_name_ref::Unknown(#unknown_var_ref {payload, class, msg_id}) => {
+                        PacketOwned::Unknown(#unknown_var_owned { payload: payload.to_vec(), class: *class, msg_id: *msg_id })
```
lkolbly, line 777: I _think_ this is the only place affected by my `alloc` comment above, correct me if I'm wrong.

lkolbly:
Thanks for the PR! I concur, it's a little weird that we have to duplicate the struct (and methods). I think we do have to do that, though: If the owned variant was a base that the ref referred to, then I don't think we could do the zero-copy thing that Ref can do (where it references some buffer that the user holds)

...that said, maybe we could do something clever with traits and default methods. You could imagine a trait like so:
```rust
pub trait SomePacket {
    fn latitude(&self) -> u32 {
        u32::from_le_bytes(self.bytes()[0..=3].try_into().unwrap())
    }
    
    fn bytes(&self) -> &[u8];
}
```
and then you could have `PacketRef` and `PacketOwned` each implement the `bytes` method, like so:
```rust
pub struct PacketRef<'a>(&'a [u8]);
pub struct PacketOwned([u8; 4]);

impl<'a> SomePacket for PacketRef<'a> {
    fn bytes(&self) -> &[u8] {
        self.0
    }
}

impl SomePacket for PacketOwned {
    fn bytes<'a>(&'a self) -> &'a [u8] {
        &self.0
    }
}
```
Which also lets you write code that's agnostic to ref or owned:
```rust
fn extract_field(bytes: impl SomePacket) -> u32 {
    bytes.latitude()
}

pub fn cvt_bytes(bytes: [u8; 4]) -> u32 {
    let p = PacketOwned(bytes);
    extract_field(p)
}
```
Which, as always, compiles [entertainingly well](https://rust.godbolt.org/z/jK9arbqM3).

Anyway, I'm sure we could go on forever down that rabbit hole. Maybe we can roll it into #43 somehow. For now though, this looks good, barring my comment about alloc.

CramBL:
IMO this should not result in duplicated disjoint packet types. A better solution is to make a packet enum with a borrowed and an owned variant, similar to how they implemented it in `egui_plot`: https://github.com/emilk/egui_plot/blob/9ebfc9c400f8835255754e6a2edb52ea77fab8c7/egui_plot/src/items/values.rs#L160

So in `egui_plot` the `PlotPoints` is analogous to the `u8` of the payload, and so the borrowed type would have a borrowed slice `&[u8]` (as it currently has) and the owned type would have an owned payload that is a `vec<u8>`.

I haven't tried implementing it, but I believe this would simply result in a `payload` enum with a borrowed and an owned variant and then the `UbxUnknownPacket` can have a `clone` or `to_owned` method that returns `Self` with an owned payload.

CramBL:
Considering just how out of date this PR is, and how many changes it contains that are **not** related to implementing an owned packet, I would rather create an issue at this point and tackled it after we're done bringing the repo up to date etc.

CramBL:
I looked a bit more into this, and there's a few challenges with going with a different approach.

## Approach of this PR

### Pros 

- The current implementation has a `PacketRef` so adding a `PacketOwned` fits the theme
- The owned variant is implemented with a fixed-size array (stack-allocated)

### Cons

- Duplicated code
- Not idiomatic to have different types named `Ref` and `Owned`

## Vec<u8> approach

### Pros

- Minimal extra code, just wraps the payload in a simple enum
- Can have a single `Packet` type (con is that it is a big breaking change)

### Cons

- `Vec<u8>` is heap allocated

## Conclusion

Considering it's a macro, the duplication is not a huge deal. The implementation looks good IMO, so maybe we just go with that? 

I didn't originally consider that the implementation was with a fixed-size array buffer, that's pretty nice imo.

CramBL:
Oh and btw, the reason the enum wrapping payload cannot be a fixed-size array is that the `UbxUnknownPacketRef` is not fixed-size, I would have to choose a payload size at compile-time. There might be a sane size to choose, but then again, is it worth the effort considering the pros and cons?

andrei-ng:
Haven't looked in detail into this. Just as a quick note, after reading your comment. We should keep the *non-heap alloc* option as having that capability is important for embedded devices support.

andrei-ng:
@CramBL , if you could bring this PR up to date after we release v0.5.0 , that would be great.

CramBL:
I've tried resolving merge conflicts with `v0.5.0` and then again with `v0.4.5` because it was so difficult, but it didn't help at all, and the formatting changes are making it even more difficult. The big problem is that since this PR the `serde` integration was implemented, so this PRs new owned/reference differentiation needs to play well with the macro derived serde integration and that is very challenging with these commits.

I would much prefer trying to port these changes manually onto the newest version, but then I would lose the commit history from @maximeborges.

IMO resolving these merge conflicts and then making it work, will be much more work than porting these changes to the tip of master.

What do you think @gwbres @andrei-ng ? If you give the go ahead then I'll start working on it.

gwbres:
@CramBL thank you so much for your time ! Yeah we should definitely start off from scratch if that branch hit a wall (too many changes). I'll try to help you both. For me this and the CfgGnss topics are the most important IMO (besides new contributions)

see you

gwbres:
> The big problem is that since this PR the serde integration was implemented, so this PRs new owned/reference differentiation needs to play well with the macro derived serde integration and that is very challenging with these commits

that makes me notice that the `serde` feature does not depend on the `std` feature, would you have `serde` require `std` activation ?

CramBL:
> > The big problem is that since this PR the serde integration was implemented, so this PRs new owned/reference differentiation needs to play well with the macro derived serde integration and that is very challenging with these commits
> 
> that makes me notice that the `serde` feature does not depend on the `std` feature, would you have `serde` require `std` activation ?

Well, serde has no_std support: https://serde.rs/no-std.html so I guess the question is if our use of serde allows no std. If it doesn't then it is an issue.

maximeborges:
No problem if you want to re-implement this feature without my commit history.
Also sorry for the rustfmt pass in the same commit 🙈
If you need any help, I might have a bit of time at the end of this week

maximeborges:
Replacing this PR by https://github.com/ublox-rs/ublox/pull/103

PR #51, opened 2022-12-01
Fix `MonVer` validation.

reitermarkus:
Fix wrong `MonVer` validation which was broken in https://github.com/lkolbly/ublox/pull/40.


reitermarkus:
Missed that there's already an open PR for this: https://github.com/lkolbly/ublox/pull/47

PR #56, opened 2023-08-23
Implement CfgValGet request and response, and NavSvIn

stevenlovegrove:
Also added some associated configuration Key enums.

I have test these changes on my F9P device but don't have any others to work with. I've been working with https://cdn.sparkfun.com/assets/f/7/4/3/5/PM-15136.pdf as the spec and I'm not aware of what other gotchas there may be, so any feedback welcome.

The Wildcard query isn't working right now because not all keys are implemented in this library - the variable length encoding requires knowledge of the length of each key, so parsing has to abort when it hits an unknown key.

Let me know if you have any comments on style etc and if this is useful!


andrei-ng:
@CramBL , this is an importan PR, we should merge this in before doing major refactoring.

andrei-ng:
@stevenlovegrove, @Hackerman342 , similar to what was done in the other PR https://github.com/ublox-rs/ublox/pull/44#issuecomment-2781598136 , we will have to make a new PR and take the changes one by one.  Unfortunately there is no way to reconcile this PR with the the state of master. Credit for the work is yours and I will try to see if I can use `cherry-pick` to get at least a few commits in, but can't promise.

andrei-ng:
Merged in via #106 and #125

PR #61, opened 2023-09-07
bump rustc and dependencies to newer versions

andrei-ng:
This PR bumps up all dependencies including rustc to 1.70.

I have been using this versions in a few applications with M8L since June and unless I missed some important reason for not bumping dependencies, I would love to see this merged in.

I update the example in `ublox_cli` to work with the new `clap` and `serialport` versions. 

@gwbres please let me know what you think.


ublox_cli/src/main.rs
```diff
@@ -78,86 +82,123 @@ impl Device {
...
                 portid: UartPortId::Uart1,
                 reserved0: 0,
                 tx_ready: 0,
                 mode: UartMode::new(DataBits::Eight, Parity::None, StopBits::One),
-                baud_rate: 9600,
-                in_proto_mask: InProtoMask::all(),
-                out_proto_mask: OutProtoMask::UBLOX,
+                baud_rate: baud,
+                in_proto_mask: InProtoMask::UBLOX,
+                out_proto_mask: OutProtoMask::union(OutProtoMask::NMEA, OutProtoMask::UBLOX),
```
gwbres, line 164: These could also benefit command line options in the future

ublox_cli/src/main.rs
```diff
@@ -78,86 +82,123 @@ impl Device {
...
+        Some("2") => SerialStopBits::Two,
+        _ => SerialStopBits::One,
     };
-    let port = serialport::open_with_settings(port, &s).unwrap();
+    let data_bits = match matches.get_one::<String>("data-bits").map(|s| s.as_str()) {
+        Some("5") => SerialDataBits::Five,
+        Some("6") => SerialDataBits::Six,
+        Some("7") => SerialDataBits::Seven,
+        _ => SerialDataBits::Eight,
+    };
```
gwbres, line 137: nice feature to have

.github/workflows/rust.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - name: Install MSRV
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.65.0
+        toolchain: 1.70.0
```
gwbres, line 26: is this bump caused by the cli::clap bump ?
As far as i'm concerned, i'm not bothered by complying to very recent compilers.
But if the library differs too much from the command line app/demo, it could be interesting that we differentiate their requirements to end users

ublox_cli/src/main.rs
andrei-ng: Good point, I can add them in a next PR. I have more coming :)

.github/workflows/rust.yml
andrei-ng: Yes indeed. That version of `clap` comes with a requirement on `1.70`. In general I also prefer using recent compilers / pacakges if there is no other restriction. 

One idea could be to add the following `rust-version = "1.70"` in the Cargo.toml manifest of `ublox_cli`.

gwbres: Yep, we need to do just that

andrei-ng: Will do later today.

gwbres (approved):

.github/workflows/rust.yml
andrei-ng: I moved the example to an `examples` folder and excluded it from the root workspace

gwbres (approved):

PR #66, opened 2023-09-16
crate won't build without any activated features

gwbres:


gwbres:
It looks like the CI does not build the crate without any options. Not sure what --no-default triggers

adrian-kong:
What is going on at the moment, the CI seems to pass just fine?

adrian-kong (approved):

gwbres:
> What is going on at the moment, the CI seems to pass just fine?

Cargo publish does not work, because its build does not pass.   
It this cargo publish invokes "cargo build --release" meaning : no features activated.
To me, it looks like the crate does not build correctly when no features are activated, but I can't reproduce in the CI

adrian-kong:
> > What is going on at the moment, the CI seems to pass just fine?
> 
> Cargo publish does not work, because its build does not pass. It this cargo publish invokes "cargo build --release" meaning : no features activated. To me, it looks like the crate does not build correctly when no features are activated, but I can't reproduce in the CI

Hmm, it seems like it ran on 1.74 nightly, 
https://docs.rs/crate/ublox/0.4.3/builds/913958
not sure if thats the issue.

andrei-ng:
> > > What is going on at the moment, the CI seems to pass just fine?
> > 
> > 
> > Cargo publish does not work, because its build does not pass. It this cargo publish invokes "cargo build --release" meaning : no features activated. To me, it looks like the crate does not build correctly when no features are activated, but I can't reproduce in the CI
> 
> Seems like ran on 1.74 nightly + docs failed, https://docs.rs/crate/ublox/0.4.3/builds/913958

I tried it locally with 1.65 and I get the same build errors with `cargo publish --dry-run` . Strange behavior from cargo...

PR #71, opened 2023-11-10
extend basic_cli example with more cmd line arguments

andrei-ng:
As discussed some time ago in https://github.com/ublox-rs/ublox/pull/61, this PR adds more CLI args via `clap` to the basic example. 

While the example of using the uBlox library itself is indeed basic, the clap argument handling has become a bit more complex. 

@gwbres let me know what you think.


gwbres:
look's great ! I think its awesome especially for people connecting to a ublox for the first time.
The only thing, I think's `ublox-cli` is a much more "intuitive" name than the proposed one right now

andrei-ng:
OK. Will change it to `ublox-cli`, I agree.

gwbres (approved):

andrei-ng:
@gwbres , change the name and also enabled NavPVT on all ports.  Thank you!

PR #79, opened 2024-09-16
Add L5 "ignore health bit" command

dgramop:
While not formally part of the F9P interface description, UBLOX modules with L5 support come with L5 "ignored" by default, since it ignores satellites with the health bit set to unhealthy.

See: https://content.u-blox.com/sites/default/files/documents/GPS-L5-configuration_AppNote_UBX-21038688.pdf

The solution is to send a configuration command, which they present as a byte string. In effect, this is just a configuration command. I added it as a value in the CfgVal enum.

### DIFF
--- a/ublox/src/ubx_packets/cfg_val.rs
+++ b/ublox/src/ubx_packets/cfg_val.rs
@@ -1094,6 +1094,10 @@ cfg_val! {
   SignalGloL1Ena,        0x10310018, bool,
   SignalGLoL2Ena,        0x1031001a, bool,
 
+  /// "Undocumented" L5 Health Bit Ignore (see
+  /// https://content.u-blox.com/sites/default/files/documents/GPS-L5-configuration_AppNote_UBX-21038688.pdf)
+  UndocumentedL5Enable,  0x10320001, bool,
+
   // CFG-TP-*
   TpPulseDef,            0x20050023, TpPulse,
   TpPulseLengthDef,      0x20050030, TpPulseLength,


gwbres:
Hello @dgramop, thank you for the proposal.

while this seems benign, I think it would really benefit being optional. For example it is not compatible with a critical application using F9P, and I have no idea how it might interfere with -10 series, for the day we inquire usage of the latest serie

dgramop:
By what mechanism would you suggest I make this optional?

Features, generics... etc?

Whatever fits your vision for the project best :)

CramBL:
I don't know how common these "undocumented" commands are, but if they're not extremely uncommon then I would gate them behind an "undocumented" feature flag, otherwise simply an "extras" feature.

andrei-ng:
> I don't know how common these "undocumented" commands are, but if they're not extremely uncommon then I would gate them behind an "undocumented" feature flag, otherwise simply an "extras" feature.

I totally agree with Marc (@CramBL ). The thing is that all that list of CfgVal's is in a macro. @CramBL do you have an advice or an idea on how to introduce such a feature flag within a macro? I failed in my simple attempts ...

CramBL:
> > I don't know how common these "undocumented" commands are, but if they're not extremely uncommon then I would gate them behind an "undocumented" feature flag, otherwise simply an "extras" feature.
> 
> I totally agree with Marc (@CramBL ). The thing is that all that list of CfgVal's is in a macro. @CramBL do you have an advice or an idea on how to introduce such a feature flag within a macro? I failed in my simple attempts ...

No concrete ideas, I'm also not a macro wizard but I have written a proc macro crate before. I'd just have to have a go at it and see if I can work out a good solution.

andrei-ng:
One more note here. These values in CfgVal list have to be manually configured by calling CfgValSetBuilder, assembling them and then sending them to the device. So although undocumented, someone needs to intentionally want to select and send this value to a device. IMO, it is ideal to have this undocumented value behind a feature flag , however, from a pragmatic point of view, if it is too complicated to do so, it would be better to merge it in and have a more explanatory comment at the top of it, rather then keeping this PR unmerged.

gwbres:
> IMO, it is ideal to have this undocumented value behind a feature flag , however, from a pragmatic point of view, if it is too complicated to do so

I agree that feature dependent stuff becomes rapidly cumbursome for the end user.
I don't see the need for a feature flag here, because `CfgValSetBuilder` allows us to stack whatever `CfgVal` we want, the end user can design its own custom configuration "rather easily".

dgramop:
Let me know if you guys need anything from me! I think I was not entirely correct to call it undocumented - it's documented on their website, just not formally listed in the interface descriptions I was able to find online. So I don't think this is "relying on a private interface" per-se... though since it isn't in the ICD they could change it, so maybe it is?

dgramop:
I also no longer have access to these ublox's

gwbres (approved):

andreinicolaegherghescu-tomtom (approved):

gwbres:
The CI of this project needs a little update and improvements, I'm OK to merge and fix the CI afterwards personnally

andrei-ng (approved):

andreinicolaegherghescu-tomtom (approved):

CramBL (approved):

PR #89, opened 2025-03-12
Create a release script

gwbres:
- a `CRATES_IO_TOKEN` is now available


gwbres:
@andrei-ng, could you write down your crates i/o username please ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: Publish
+        env:
+          TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
+        run: |
+          cargo login $TOKEN
+          cargo publish --allow-dirty -p ublox_derive
```
CramBL, line 31: Why is `--allow-dirty` needed?

CramBL, line 31:
```suggestion
      - name: Publish
        run: cargo publish --token ${{ secrets.CRATES_IO_TOKEN }} --allow-dirty -p ublox_derive
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: Publish
+        env:
+          TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
+        run: |
+          cargo login $TOKEN
+          cargo publish --allow-dirty -p ublox
```
CramBL, line 52:
```suggestion
      - name: Publish
        run: cargo publish --token ${{ secrets.CRATES_IO_TOKEN }}  --allow-dirty -p ublox
```

andrei-ng:
> @andrei-ng, could you write down your crates i/o username please ?

It is the same as my GitHub username: andrei-ng

CramBL (changes requested):
Great! just have a few comments and suggestions to simplify

.github/workflows/release.yml
gwbres: In practice it is not, I copied from somewhere else.
Publication might fail for example, if your CI generates stuff that is not _gitignored_, the branch becomes "dirty" and publication is disabled. To avoid this, I simply put it all the time, I consider that tagging demonstrates that CI and release should be obtained

andrei-ng (approved):

gwbres:
> It is the same as my GitHub username: andrei-ng

you should have received an invite, 
so you guys are not dependent on myself form now on,
also the release script now exists and a simple tag will upload to the portal

andrei-ng:
> > It is the same as my GitHub username: andrei-ng
> 
> you should have received an invite, so you guys are not dependent on myself form now on, also the release script now exists and a simple tag will upload to the portal

I have and acccepted it. Thank you for taking care of this!

PR #96, opened 2025-03-14
Update README.md

CramBL:
Readme polish, mostly just syntax highlighting for examples. We should also test the examples though...

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,8 @@ Constructing Packets
 ====================
 
 Constructing packets happens using the `Builder` variant of the packet, for example:
-```
+
+```rust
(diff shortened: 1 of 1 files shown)


gwbres:
> We should also test the examples though...

not sure about that though, in the end it needs to connect to the hardware. Do you have something in mind ?

andrei-ng (approved):

CramBL:
> > We should also test the examples though...
> 
> not sure about that though, in the end it needs to connect to the hardware. Do you have something in mind ?

I just mean compile it

gwbres:
> I just mean compile it

I think cargo test does compile all integrated examples

PR #100, opened 2025-03-27
Fix CI release script

gwbres:
* Ublox-rs needs ublox_derive to be released first
  * continue-on-error: permits ublox_derive publication failure and ublox publication success, when only releasing the latter
  * ublox auto-docs is currently in failure, not sure why it is not picked up yet by CI. I introduce a job dedicated to lib documentations

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   publish_derive:
-    name: Publish
+    name: Publish derive
     runs-on: ubuntu-latest
     continue-on-error: true
     if: github.ref_type == 'tag'
@@ -27,7 +27,7 @@ jobs:
         run: cargo publish --token ${{ secrets.CRATES_IO_TOKEN }} -p ublox_derive
 
   publish_ublox:
-    name: Publish
+    name: Publish ublox
     needs: ['publish_derive']
     runs-on: ubuntu-latest
     continue-on-error: true
@@ -41,7 +41,7 @@ jobs:
         with:
           toolchain: stable
 
-      - name: Publish
+      - name: Publish 
         run: cargo publish --token ${{ secrets.CRATES_IO_TOKEN }} -p ublox
   
   release:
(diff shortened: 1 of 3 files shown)


ublox/Cargo.toml
```diff
@@ -18,6 +18,10 @@ default = ["std", "serde", "ubx_proto23"]
 alloc = []
 std = []
 
+[package.metadata.docs.rs]
+all-features = false
+rustdoc-args = ["--cfg", "docrs", "--generate-link-to-definition"]
+
```
gwbres, line 24: This prevailed publishing a new library as of right now

CramBL (approved):
Looks good to me. I don't understand if there's some issue with building the documentation?

You could use the "working-directory" key instead of `cd` but that's up to you

.github/workflows/release.yml
```diff
@@ -28,6 +28,7 @@ jobs:
 
   publish_ublox:
     name: Publish
```
andrei-ng, line 30: Could you change this to `Publish ublox_derive` or `Publish derive`, to make it clearer in the UI which step it is

andrei-ng (approved):

andrei-ng:
Left one comment regarding a cosmetic naming of the CI jobs. Besides that, looks good.

gwbres:
thank you all

cargo workspace is flawed and terrible, I'm pretty sure these tools are maintained by developpers, not engineers nor scientists.
I will create one last fix, I'm not spending my time on crap like that, I have much better things to do

andrei-ng:
Thank you for taking care of the release! :fireworks: :bowing_man:

gwbres:
> @gwbres Thank you for taking care of the release! 🎆 🙇‍♂️

@andrei-ng 👍 

I can't seem to find a correct way to integrate the Readme file in the src/lib.rs file. I've had cargo workspaces in the past and don't recall any issue. The CI/CD documentation job did pass, but for some reason, the documentation run that "publishing" will trigger, does not.. Well. We have more important subjects to focus on, let's just keep it in mind

PR #105, opened 2025-04-15
fix dds example and refactor naming and Cargo.toml

andrei-ng:


CramBL (approved):
Looks good but why don't we use a single workspace?

We could use the `example` cargo.toml manifest key: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples

Or do it like `ratatui` and many others do where they are just members of the workspace: https://github.com/ratatui/ratatui/blob/02e53de0f82bc2364f2ba4345cc5a635347e3eb4/Cargo.toml

andrei-ng:
Very good you bring this up. The reason ... is ... long story short ...legacy. I will move them to the canonical place and rebase.

CramBL:
When we have a single workspace there is no longer a need to `cd` into the examples directory, and some of the CI jobs should be possible to merge now

andrei-ng:
Merged the MSRV into a single CI job. 
I left the others separate as the examples require extra dependencies which I don't want to build by default.

CramBL:
> Merged the MSRV into a single CI job. I left the others separate as the examples require extra dependencies which I don't want to build by default.

makes sense

PR #73, opened 2023-11-16
Add NAV-SIG and CFG-GNSS messages

aholtzma-am:
Another couple of messages to add.

PR #106, opened 2025-04-18
Add CfgValGet message for requesting key values.

andrei-ng:
Porting and splitting the additions from PR #56 into this and separate PRs as our master has diverged substantially with the introduction of the feature flags. Credit for the work goes to @stevenlovegrove, @Hackerman342


CramBL:
You are still working on making the semver checks pass right? Just checking because I prefer waiting for CI to pass before starting review

andrei-ng:
> You are still working on making the semver checks pass right? Just checking because I prefer waiting for CI to pass before starting review

Semver won't pass as this PR introudces 
 - new packet in Recv enum
 - renames enum variants in CfgVal

So this will require a new major bump eventually :(

CramBL:
> > You are still working on making the semver checks pass right? Just checking because I prefer waiting for CI to pass before starting review
> 
> Semver won't pass as this PR introudces
> 
>     * new packet in Recv enum
> 
>     * renames enum variants in CfgVal
> 
> 
> So this will require a new major bump eventually :(

Oh okay. I'll take a look then, maybe we should consider 1.0.0 in the not too distant future.

ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -42,6 +42,16 @@ impl KeyId {
     pub const fn item_id(&self) -> u8 {
         self.0 as u8
     }
+
+    pub fn extend_to<T>(&self, buf: &mut T) -> usize
```
CramBL, line 46: Is now a good time to start more rigorously documenting the API? 😄 

```suggestion
    /// Extend a buffer with the contents of [KeyId].
    pub fn extend_to<T>(&self, buf: &mut T) -> usize
```

ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -236,15 +295,19 @@ macro_rules! cfg_val {
...
-              Self::$cfg_item(from_cfg_v_bytes!(&buf[4..], $cfg_value_type))
+              Some(Self::$cfg_item(from_cfg_v_bytes!(&buf[4..], $cfg_value_type)))
             },
           )*
-          _ => unimplemented!("unknown key ID: 0x{:8X}", key_id),
+          _ => {
+            // TODO: add a mechanism to log such messages that supports also no_std
+            // eprintln!("unknown key ID: 0x{:8X}", key_id);
+            None
+           },
```
CramBL, line 310: This suggest that the function should actually return `Result` instead of `Option`. If you have an unknown key ID isn't that just an error?

ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -1118,9 +1265,28 @@ pub enum TpPulse {
...
     /// Time pulse length
     Length = 1,
 }
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum TModeMode {
+    Disabled = 0,
+    SurveyIn = 1,
+    // true ARP position information required
```
CramBL, line 1281:
```suggestion
    /// true ARP position information required
```

ublox/src/ubx_packets/packets.rs
```diff
@@ -2358,6 +2358,97 @@ struct CfgValSet<'a> {
...
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+struct CfgValGet<'a> {
+    /// Message version
+    version: u8,
+    /// The layers from which the configuration items should be retrieved
+    #[ubx(map_type = CfgReadLayer)]
+    layers: u8,
+    position: u16,
+    cfg_keys: &'a [CfgKey],
```
CramBL, line 2377: Just to get some kind of association between the lifetime name and the value that determines the lifetime

```suggestion
struct CfgValGet<'k> {
    /// Message version
    version: u8,
    /// The layers from which the configuration items should be retrieved
    #[ubx(map_type = CfgReadLayer)]
    layers: u8,
    position: u16,
    cfg_keys: &'k [CfgKey],
```

ublox/src/ubx_packets/packets.rs
```diff
@@ -2358,6 +2358,97 @@ struct CfgValSet<'a> {
...
+    data: &'a [u8],
+}
+
+impl<'a> CfgValReadIter<'a> {
+    fn new(data: &'a [u8]) -> Self {
+        // todo!("stolen from RxmRawxInfoIter. Reimplement.");
+        Self { data }
+    }
+
+    fn is_valid(bytes: &'a [u8]) -> bool {
```
CramBL, line 2430: This lifetime should not be needed

```suggestion
    fn is_valid(bytes: &[u8]) -> bool {
```

ublox/src/ubx_packets/packets.rs
```diff
@@ -2358,6 +2358,97 @@ struct CfgValSet<'a> {
...
+    type Item = CfgVal;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if Self::is_valid(self.data) {
+            if let Some(cfg_val) = CfgVal::parse(self.data) {
+                self.data = &self.data[cfg_val.len()..];
+                return Some(cfg_val);
+            }
+            // TODO: Is there some logging mechanism?
+            // eprintln!("Failure parsing key in (key,value) list, {:?}", self.data);
```
CramBL, line 2446: We could use https://defmt.ferrous-systems.com.

Then we could wrap them in a macro and support the normal std logging facade as well.

ublox/src/ubx_packets/packets.rs
```diff
@@ -2379,19 +2470,17 @@ impl<'a> CfgValIter<'a> {
     }
 }
 
-impl core::iter::Iterator for CfgValIter<'_> {
```
CramBL, line 2382: Pretty sure this lifetime can be elided

ublox/src/ubx_packets/cfg_val.rs
CramBL, line 46: To be honest I don't really understand this implementation, why do we require the buffer to have the same length as what we extend it with?

CramBL (changes requested):
Have a bunch of comments, some are a little more general, it's minor things overall (I hope).

gwbres:
Hello, I'll have some time to review your work and test it tomorrow, I will keep you updated

ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -120,6 +136,21 @@ macro_rules! from_cfg_v_bytes {
             _ => unreachable!(),
         }
     };
+    ($buf:expr, TModeMode) => {
+        match $buf[0] {
+            0 => TModeMode::Disabled,
+            1 => TModeMode::SurveyIn,
+            2 => TModeMode::FixedMode,
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
andrei-ng: Good point, thanks!
I added specific messages to almost all `unreachable` code blocks.

ublox/src/ubx_packets/packets.rs
andrei-ng: Not sure it matters that much here as we have a single lifetime specifier. But I will do it.

ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -42,6 +42,16 @@ impl KeyId {
     pub const fn item_id(&self) -> u8 {
         self.0 as u8
     }
+
+    pub fn extend_to<T>(&self, buf: &mut T) -> usize
+    where
+        T: core::iter::Extend<u8>,
+    {
+        let bytes = self.0.to_le_bytes();
+        assert_eq!(bytes.len(), Self::SIZE);
```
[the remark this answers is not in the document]
andrei-ng: True. It should not be there. Probably it was a leftover from the debugging phase of the original development.

ublox/src/ubx_packets/cfg_val.rs
andrei-ng: true

ublox/src/ubx_packets/packets.rs
andrei-ng: Interestingly, trying to fix your suggestion made me catch this ... The lifetime needs to always be `'a` as it is hardcoded in the macro here https://github.com/ublox-rs/ublox/blob/0f9889bf9fc30ef71e13cff6610c20e2bf8beb6d/ublox_derive/src/output.rs#L449

ublox/src/ubx_packets/packets.rs
andrei-ng: I will make an issue out of this. We should add something along those lines, but not as part of this PR.

ublox/src/ubx_packets/cfg_val.rs
andrei-ng: Looking at the code, this seems to be a mistake. Maybe the intention was to limit the number of keys the user can request. The F9R documentation specifies that `This message is limited to containing a maximum of 64 key IDs.` If I were to guess, maybe the intention was to limit the buffer to `MAX_CFG_KEYS = 64`. That is something @stevenlovegrove could confirm/deny.

andrei-ng: Anyway, in the meantime I've decided to remove the `asser` alltogether.

ublox/src/ubx_packets/cfg_val.rs
andrei-ng: I think you are right. The `parse` function seems to be called only in the `CfgValIter` `next` method so at that level, Result can be handled the same as the Option.

CramBL: good point but a little confusing...

ublox/src/ubx_packets/packets.rs
CramBL: haha yea that's pretty ugly, it has to be part of the macro refactor at a later time imo

CramBL:
> @CramBL , @gwbres, originally when I created the PR, I simply cherry-picked the old PR's commits and fixed any compiler issues along the way, As @CramBL proposed improvements, I looked closer at the implementation and I realised that this implemenation brought back an older issues I had with the macros implementation in the crate.
> 
> The messages added by this PR `CfgValGet` and `CfgValGetResponse` are in implemented using the `#[ubx_packet_send]` macro and `#[ubx_packet_recv]` respectively. But these are in principle the same message (same class and same message ID), and our crate has in theory support for `#[ubx_packet_recv_send]` messages. However, this macro never worked with messages that have variable payloads. I faced the same issue way back when I was trying to implement the `UbxEsfMeas` message which can also be sent or received.
> 
> So my question to both of you is how to address this going forward. For this kind of messages, do we stick with two different derived implementations: one using `#[ubx_packet_send]` macro and one with `#[ubx_packet_recv]` or should we spend some time and try to find a way to make `#[ubx_packet_recv_send]` work in this kind of cases. (for that I would need @CramBL macro know-how though ... ) . And how do we go forward with this PR?

I guess we **should** try to make `#[ubx_packet_recv_send]` work.

Should we start by getting all of this stuff merged in good condition and then make `#[ubx_packet_recv_send]` work afterwards? We need to do it before the next release though, so we don't immediately have to do another breaking change.

ublox/src/ubx_packets/packets.rs
andrei-ng: I'll turn this into an issue and resolve the comment.

andrei-ng:
> Should we start by getting all of this stuff merged in good condition and then make `#[ubx_packet_recv_send]` work afterwards? We need to do it before the next release though, so we don't immediately have to do another breaking change.

Yes, might thinking would be the same, 
 - we **should** first try to make it work, if we fail, then we know we've tried and we don't have a clear way forward.
 - I also think we should merge this the rest of the work from the old PR in and only then rework  `#[ubx_packet_recv_send]` prior to any major release. 
 - the fact that  `#[ubx_packet_recv_send]` is not working with variable payload is bloking me in implementing `EsfMeas` and as far as I remember blocking at least a few others, like @aholtzma-am for CfgGnss (and @gwbres ) [as discussed  [here](https://github.com/ublox-rs/ublox/pull/73#issuecomment-2729664598), so it is an issue that comes up again and again :grin:

CramBL (approved):
Looks great!

gwbres (approved):

gwbres:
So you manage to make it work ? I would recommend setting a maximal number of item, that should work and remains simple.
It looks like what you did there 🎉

ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -1124,3 +1244,19 @@ pub enum TpPulseLength {
     /// Time pulse length
     Length = 1,
 }
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum TModeMode {
+    Disabled = 0,
+    SurveyIn = 1,
+    // true ARP position information required
```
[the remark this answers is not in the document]
andrei-ng: Done!

ublox/src/ubx_packets/packets.rs
```diff
@@ -5069,6 +5085,24 @@ struct SecUniqId {
...
 }
 
+/// The `CfgReadLayer` enum is used to specify the configuration layer to read from.
+/// The configuration system in the ublox device is stacked, so a property
+/// may be empty for a particular layer and you will receive a NAK.
+#[ubx_extend]
+#[ubx(from_unchecked, into_raw, rest_error)]
+#[repr(u8)]
+#[derive(Debug, Copy, Clone)]
+pub enum CfgReadLayer {
```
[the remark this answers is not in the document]
andrei-ng: Unfortunately it is not possible. uBlox must have messed it up here. The definition of the Layer is different between CFG-VALSET and CFG-VALGET, so unfortunately we cannot reuse it. I just renamed the two to give them a more representative name `CfgLayerSet` and `CfgLayerGet`. Other suggestions are welcome :)

PR #107, opened 2025-04-18
Make examples cargo uniform and add dependabot

andrei-ng:


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: daily
```
CramBL, line 6: Are we sure we want daily? Even weekly can be a bit much.

Do we want to group PRs, allowing multiple cargo updates per dependabot PR?

Do we make changelog entries for dependency updates? 
We could choose not to, but we would definitely be motivated to do a release on occasion just based on a bunch of dependency updates or a particular valuable fix in a dependency, then we could just have "dependency updates" in the changelog, but that's not so helpful.

I have an action that makes changelog entries based on dependabot PRs, we could use it, it also makes some trade-offs. I'm willing to update it if we want to tweak the approach though.

CramBL (approved):
I approve but we might make an issue or something out of the comment I have and resolve it in a discussion + future PRs.

.github/dependabot.yml
andrei-ng: Daily indeed, might be a bit too much. I will change it to weekly. 

> Do we want to group PRs, allowing multiple cargo updates per dependabot PR?
Not sure what to say there. I don't have experience with that approach. It might work better as we don't have many dependencies. But it might shoot us in the face if one dependency causes issues, then we will have to update that PR and exclude the troublesome one. If from your experience it worked, better, let's try it (we can always change it later)

>Do we make changelog entries for dependency updates?
Tricky one and hot topic I guess. I've seen this come up on other crates as well. I think we should make a new release if important dependencies are updated.  And if we end up making releases, then I guess Changelog needs to be updated to maintain some nice history besides git log.

> I have an action that makes changelog entries based on dependabot PRs, we could use it, it also makes some trade-offs. I'm willing to update it if we want to tweak the approach though.

Yes, give it a try. Feel free to push to this PR on top of my commits. I will only swap `daily` to `weekly` and let you updated it further.

andrei-ng:
> I approve but we might make an issue or something out of the comment I have and resolve it in a discussion + future PRs.

I just saw this after I left my comment. Then I will merge this as is and we tackle your comment in an issue + future PRs

Issue #114, opened 2025-04-18
CI improvements - Typo checking


CramBL:
Nothing worse than typos to ruin good documentation and code readability:

Proposed solution: https://github.com/crate-ci/typos

PR #118, opened 2025-04-19
fix typos in ublox and ublox_derive

CramBL:
Preparation for https://github.com/ublox-rs/ublox/issues/114 :smile: 

Fix some typos discovered via [typos](https://github.com/crate-ci/typos) mostly.

Also fix one instance of `i.e.` when author meant `e.g.`

The typos were mostly in comments but a few were in code, which causes breaking changes, I've documented those breaking changes in the changelog

### DIFF
--- a/ublox/src/ubx_packets/packets.rs
+++ b/ublox/src/ubx_packets/packets.rs
@@ -403,7 +403,7 @@ impl NavPvtFlags3 {
         self.invalid_llh
     }
 
-    /// F9R interface descriptions document specifies that this byte is unused
+    /// F9R interface description document specifies that this byte is unused
     #[cfg(feature = "ubx_proto23")]
     pub fn age_differential_correction(&self) -> u8 {
         self.age_differential_correction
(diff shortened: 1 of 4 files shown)


andrei-ng (approved):

ublox/src/ubx_packets/packets.rs
```diff
@@ -403,7 +403,7 @@ impl NavPvtFlags3 {
         self.invalid_llh
     }
 
-    /// F9R interface descritpion document specifies that this byte is unused
+    /// F9R interface descriptions document specifies that this byte is unused
```
gwbres, line 406:
```suggestion
    /// F9R interface description document specifies that this byte is unused
```

gwbres (approved):

ublox/src/ubx_packets/packets.rs
andrei-ng: Sorry @gwbres , I missed that. Good catch.

CramBL: same error on line 418

Issue #123, opened 2025-04-22
Make `#[ubx_packet_recv_send]` work with variable length payload


andrei-ng:
As discussed [here](https://github.com/ublox-rs/ublox/pull/106#issuecomment-2821491843)

andrei-ng:
- CFG-GNSS done.
- CFG-VAL - won't do; distinct types for send vs recv
- Next in line ESF-MEAS

PR #124, opened 2025-04-23
wip: make CfgValGet send and recv

andrei-ng:
Relates to #123

### DIFF
--- a/ublox/src/ubx_packets/packets/cfg_val.rs
+++ b/ublox/src/ubx_packets/packets/cfg_val.rs
@@ -10,11 +10,15 @@ use crate::FieldIter;
 use alloc::vec::Vec;
 
 use crate::{
-    cfg_val::{CfgKey, CfgVal},
+    cfg_val::CfgVal,
+    // cfg_val::CfgKey,
     error::ParserError,
-    ubx_checksum, UbxPacketMeta, SYNC_CHAR_1, SYNC_CHAR_2,
+    ubx_checksum,
+    UbxPacketMeta,
+    SYNC_CHAR_1,
+    SYNC_CHAR_2,
 };
-use ublox_derive::{ubx_extend, ubx_extend_bitflags, ubx_packet_recv, ubx_packet_send};
+use ublox_derive::{ubx_extend, ubx_extend_bitflags, ubx_packet_recv_send, ubx_packet_send};
 
 #[ubx_packet_send]
 #[ubx(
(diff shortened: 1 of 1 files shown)


andrei-ng:
@CramBL, the thing about `ubx_packet_recv_send` is that I think it was never implemented with payloads with variable fields in mind. 

- `ubx_packet_send` macro works fine with fields that use references and lifetimes 
- `ubx_packet_recv` macro works fine with packets with variable fields and require you to implement an iterator

But having `ubx_packet_recv_send` with a payload with variable fields seems to break when one needs to implement the iterator.  I am looking into the macros, but I am stuck :cry:

CramBL:
> @CramBL, the thing about `ubx_packet_recv_send` is that I think it was never implemented with payloads with variable fields in mind.
> 
>     * `ubx_packet_send` macro works fine with fields that use references and lifetimes
> 
>     * `ubx_packet_recv` macro works fine with packets with variable fields and require you to implement an iterator
> 
> 
> But having `ubx_packet_recv_send` with a payload with variable fields seems to break when one needs to implement the iterator. I am looking into the macros, but I am stuck 😢

Yea looks pretty rough... I'll have a look at it later this week I expect.

CramBL:
There's a lot to unpack here, so what I'm actually doing right now is doing major refactors to the macro crate. I'm basically just dividing stuff into smaller files, but to solve this issue I think the macro crate needs to handle generics, it needs to do that anyway to be able to name lifetimes (I already solved that).

WIP but this is gonna be a massive refactor FYI.

CramBL:
@andrei-ng I made it compile but it's kind of terrible. I mean, it's a hack. 

I just check if a getter of a variable sized packet has a lifetime, and in that case I remove all the generic tokens so I remove `<'a>` for `CfgValIter<'a>`

Can you look at the actual `CfgValGet` implementation again? What more is needed to land this?

CramBL:
I added this new `#[debug_this]` attribute:

```rust
#[proc_macro_attribute]
pub fn debug_this(
    _attr: proc_macro::TokenStream,
    input: proc_macro::TokenStream,
) -> proc_macro::TokenStream {
    let mut input = parse_macro_input!(input as DeriveInput);
    let debug_attr: syn::Attribute = syn::parse_quote! {
        #[debug_this]
    };
    input.attrs.push(debug_attr);
    input.to_token_stream().into()
}
```

You can add it to a struct and then the `DebugContext` will print macro context for that specific struct e.g.

```rust
dbg_ctx.print_at(
        file!(),
        line!(),
        format_args!("variable_size_field: out_ty:\n{:?}", out_ty),
    );
```

It is not printed in a pretty way, but I just copied it from the terminal into a file formatted by rustfmt, which was pretty powerful even though it was very primitive.

If we used `syn` v2 we would have access to more powerful macro logging tools.

CramBL:
Maybe I should make a separate PR for the macro changes?

CramBL:
I cherry-picked all my commits that touch the derive crate and made this PR: https://github.com/ublox-rs/ublox/pull/130

andrei-ng:
> Can you look at the actual CfgValGet implementation again? What more is needed to land this?

@CramBL, actually ... I just realised that this one might be a bit more tricky to get right in one go.
The docs specifiy that when sending `CfgValGet`, the payload is a repeated list of `CfgKey`s which are 4 byte in length. When getting the response, the payload contains a repeated block of 1 byte of `CfgData` . I am not sure if we can achieve this without additional macro changes on top of the ones you've already made.

After you merge #130 I can look whether your implemenation fixes first `CFG-GNSS` for ublox M8* as that is much easier to approach.

CramBL:
> @CramBL, actually ... I just realised that this one might be a bit more tricky to get right in one go. The docs specifiy that when sending `CfgValGet`, the payload is a repeated list of `CfgKey`s which are 4 byte in length. When getting the response, the payload contains a repeated block of 1 byte of `CfgData` . I am not sure if we can achieve this without additional macro changes on top of the ones you've already made.

Okay I'll look at it once you have an idea of how it could look.

> After you merge #130 I can look whether your implemenation fixes first `CFG-GNSS` for ublox M8* as that is much easier to approach.

Cool. I merged it now and updated this branch.

andrei-ng:
> > @CramBL, actually ... I just realised that this one might be a bit more tricky to get right in one go. The docs specifiy that when sending `CfgValGet`, the payload is a repeated list of `CfgKey`s which are 4 byte in length. When getting the response, the payload contains a repeated block of 1 byte of `CfgData` . I am not sure if we can achieve this without additional macro changes on top of the ones you've already made.
> 
> Okay I'll look at it once you have an idea of how it could look.
> 
> > After you merge #130 I can look whether your implemenation fixes first `CFG-GNSS` for ublox M8* as that is much easier to approach.
> 
> Cool. I merged it now and updated this branch.

I will help. I propose 
 1. make a PR with CFG-GNSS and get that working
 2. make a PR with ESF-MEAS and get that working: in the past this was also opening another can of worms as the macro only allows to have expandable blocks as the last field, but this message has one extra field after the expandable blocks
 3. get back to this PR and see if there is a way to reconcile this ...

andrei-ng:
> I don't understand what "expandable blocks" are. What blocks? Keep in mind, as silly as it is, I still have not really done any work on a ublox device (the project has been post-poned several times at this point) at work, so I'm still a complete ublox noob.

Maybe "expandable" was not the right choice of words, I should have better said, repetable blocks. 
All these 3 messages have a repetable block `N` times within the payload, that is why the crate has the implementation with the `Iter` over the respective field. However, `ESF-MEAS` has an extra "optional" field after the repetable blocks 

image

Our derive crate does not support this extra field when the packet is `send` type: https://github.com/ublox-rs/ublox/blob/1d229af7228ef5d74d04bfd9879a24edb07a819a/ublox_derive/src/output/gen_send_code.rs#L42-L44

CramBL:
> Maybe "expandable" was not the right choice of words, I should have better said, repetable blocks. All these 3 messages have a repetable block `N` times within the payload, that is why the crate has the implementation with the `Iter` over the respective field. However, `ESF-MEAS` has an extra "optional" field after the repetable blocks
> 
> image
> 
> Our derive crate does not support this extra field when the packet is `send` type:
> 
> https://github.com/ublox-rs/ublox/blob/1d229af7228ef5d74d04bfd9879a24edb07a819a/ublox_derive/src/output/gen_send_code.rs#L42-L44

Okay, I see. So how do you know if the optional message appears or not? Is there a payload length field that will indicate it or something?

If you implement it "how it's supposed to work" then I can adapt the macro to support it.

andrei-ng:
> > Maybe "expandable" was not the right choice of words, I should have better said, repetable blocks. All these 3 messages have a repetable block `N` times within the payload, that is why the crate has the implementation with the `Iter` over the respective field. However, `ESF-MEAS` has an extra "optional" field after the repetable blocks
> > image
> > Our derive crate does not support this extra field when the packet is `send` type:
> > https://github.com/ublox-rs/ublox/blob/1d229af7228ef5d74d04bfd9879a24edb07a819a/ublox_derive/src/output/gen_send_code.rs#L42-L44
> 
> Okay, I see. So how do you know if the optional message appears or not? Is there a payload length field that will indicate it or something?
> 
> If you implement it "how it's supposed to work" then I can adapt the macro to support it.

OK. Will make a PR with how I think it might need to work.

andrei-ng:
One we fix #134, I will extend the example added there to build, send and recv the `CfgValGet` message.

CramBL:
> @CramBL , I spent some time last evening looking into this specific attempt. I think there is no point to try to generalize `recv_send` macro for this use case. CfgVal Send/Recv are two distinct types. They have entirely different fields. One has a dedicated field for `CfgKey`, the recv type contains key-value pairs.
> 
> I think our efforts are better spent in other parts of the crate.
> 
> The only thing I would do, I would rename the structs from `CfgValGetSend` and `CfgValGetRecv` to `CfgValGetRequest` and `CfgValGetResponse`. I will do that in a different PR as this is too old and out of sync.

Makes sense!

PR #125, opened 2025-04-23
Port extraction of RTCM packets from buffer (PR #56)

andrei-ng:
Port this commit [835f44e](https://github.com/ublox-rs/ublox/pull/56/commits/835f44e821b8a4c73d155e365ce52cd363482ca0) from https://github.com/ublox-rs/ublox/pull/56

This implementation allows for extracting RTCM packets from the captured data by this crate.

It also completes the porting of #56  as any other packet additions were added in the meantime since the creation of that PR.

### DIFF
--- a/ublox/src/parser.rs
+++ b/ublox/src/parser.rs
@@ -9,6 +9,8 @@ use crate::{
     },
 };
 
+pub(crate) const RTCM_SYNC_CHAR: u8 = 0xd3;
+
 /// This trait represents an underlying buffer used for the Parser. We provide
 /// implementations for `Vec<u8>` and for `FixedLinearBuffer`, if you want to
 /// use your own struct as an underlying buffer you can implement this trait.
@@ -191,7 +193,7 @@ impl<T: UnderlyingBuffer> Parser<T> {
         self.buf.len()
     }
 
-    pub fn consume<'a>(&'a mut self, new_data: &'a [u8]) -> ParserIter<'a, T> {
+    pub fn consume_ubx<'a>(&'a mut self, new_data: &'a [u8]) -> UbxParserIter<'a, T> {
         let mut buf = DualBuffer::new(&mut self.buf, new_data);
 
         for i in 0..buf.len() {
@@ -201,7 +203,21 @@ impl<T: UnderlyingBuffer> Parser<T> {
             }
         }
 
-        ParserIter { buf }
+        UbxParserIter { buf }
+    }
+
+    pub fn consume_ubx_rtcm<'a>(&'a mut self, new_data: &'a [u8]) -> UbxRtcmParserIter<'a, T> {
+        let mut buf = DualBuffer::new(&mut self.buf, new_data);
+
+        for i in 0..buf.len() {
+
+            if buf[i] == SYNC_CHAR_1 || buf[i] == RTCM_SYNC_CHAR {
+                buf.drain(i);
+                break;
+            }
+        }
+
+        UbxRtcmParserIter { buf }
     }
 }
 
@@ -415,18 +431,13 @@ impl UbxChecksumCalc {
 }
 
 /// Iterator over data stored in `Parser` buffer
-pub struct ParserIter<'a, T: UnderlyingBuffer> {
+pub struct UbxParserIter<'a, T: UnderlyingBuffer> {
     buf: DualBuffer<'a, T>,
 }
 
-impl<T: UnderlyingBuffer> ParserIter<'_, T> {
-    fn find_sync(&self) -> Option<usize> {
-        (0..self.buf.len()).find(|&i| self.buf[i] == SYNC_CHAR_1)
-    }
-
-    fn extract_packet(&mut self, pack_len: usize) -> Option<Result<PacketRef, ParserError>> {
-        if !self.buf.can_drain_and_take(6, pack_len + 2) {
-            if self.buf.potential_lost_bytes() > 0 {
+fn extract_packet_ubx<'a, 'b, T: UnderlyingBuffer>(buf : &'b mut DualBuffer<'a, T>, pack_len: usize) -> Option<Result<PacketRef<'b>, ParserError>> {
+    if !buf.can_drain_and_take(6, pack_len + 2) {
+        if buf.potential_lost_bytes() > 0 {
                 // We ran out of space, drop this packet and move on
                 self.buf.drain(2);
                 return Some(Err(ParserError::OutOfMemory {
@@ -456,13 +467,23 @@ impl<T: UnderlyingBuffer> ParserIter<'_, T> {
             Ok(x) => x,
             Err(e) => {
                 return Some(Err(e));
-            },
+            }
         };
         return Some(match_packet(
             class_id,
             msg_id,
             &msg_data[..msg_data.len() - 2], // Exclude the checksum
         ));
+}
+
+impl<'a, T: UnderlyingBuffer> UbxParserIter<'a, T> {
+    fn find_sync(&self) -> Option<usize> {
+        for i in 0..self.buf.len() {
+            if self.buf[i] == SYNC_CHAR_1 {
+                return Some(i);
+            }
+        }
+        None
     }
 
     #[allow(clippy::should_implement_trait)]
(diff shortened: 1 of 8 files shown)


gwbres:
Hello, please do not implement RTCM logic and use https://docs.rs/rtcm-rs/latest/rtcm_rs/ instead, which already exists and supports `no-std`. So your logic should stop at extracting the bytes. Proposing an optional bridge (with `rtcm` feature might be useful)

andrei-ng:
> Hello, please do not implement RTCM logic and use https://docs.rs/rtcm-rs/latest/rtcm_rs/ instead, which already exists and supports `no-std`. So your logic should stop at extracting the bytes. Proposing an optional bridge (with `rtcm` feature might be useful)

Totally agree. This commit only allows for extraction of the data from the buffer. For parsing the bytes into actual RTCM messages we should indeed use the RTCM-RS crate. 
We should also add a similar buffer parsing function as was done in this PR and any NMEA crate for extracting the NMEA messages.

gwbres:
I don't this belongs to this library, for the simple reason that it is already complex enough, and people interested in parsing RTCM from their ublox have another possible scenario (sample applies to NMEA). They can simply activate that stream using the UBX library, and switch over to RTCM listening. And the dual case RTCM/UBX can be handled externally. Is there any advantage this proposal would have over everything handled externally ?

ublox/src/parser.rs
```diff
@@ -414,55 +427,71 @@ impl UbxChecksumCalc {
...
-        if (ck_a, ck_b) != (expect_ck_a, expect_ck_b) {
-            self.buf.drain(2);
-            return Some(Err(ParserError::InvalidChecksum {
-                expect: u16::from_le_bytes([expect_ck_a, expect_ck_b]),
-                got: u16::from_le_bytes([ck_a, ck_b]),
+fn extract_packet_ubx<'b, T: UnderlyingBuffer>(
+    buf: &'b mut DualBuffer<'_, T>,
+    pack_len: usize,
+) -> Option<Result<PacketRef<'b>, ParserError>> {
+    if !buf.can_drain_and_take(6, pack_len + 2) {
```
CramBL, line 452: Is it possible to name these magic numbers?

ublox/src/parser.rs
```diff
@@ -414,55 +427,71 @@ impl UbxChecksumCalc {
...
-        };
-        return Some(match_packet(
-            class_id,
-            msg_id,
-            &msg_data[..msg_data.len() - 2], // Exclude the checksum
-        ));
+        return None;
+    }
+    let mut checksummer = UbxChecksumCalc::new();
+    let (a, b) = buf.peek_raw(2..(4 + pack_len + 2));
```
CramBL, line 463: Would be nice to replace `4 + pack_len + 2` with a variable that tells readers what it means

ublox/src/parser.rs
```diff
@@ -414,55 +427,71 @@ impl UbxChecksumCalc {
...
+    checksummer.update(a);
+    checksummer.update(b);
+    let (ck_a, ck_b) = checksummer.result();
+
+    let (expect_ck_a, expect_ck_b) = (buf[6 + pack_len], buf[6 + pack_len + 1]);
```
CramBL, line 468: Also some magic numbers here

ublox/src/parser.rs
```diff
@@ -496,7 +525,109 @@ impl<T: UnderlyingBuffer> ParserIter<'_, T> {
...
+            if self.buf[i] == RTCM_SYNC_CHAR {
+                return NextSync::Rtcm(i);
+            }
+        }
+        NextSync::None
+    }
+
+    #[allow(clippy::should_implement_trait)]
+    /// Analog of `core::iter::Iterator::next`, should be switched to
+    /// trait implementation after merge of https://github.com/rust-lang/rust/issues/44265
```
CramBL, line 582: The linked issue has been resolved btw 😄

ublox/src/parser.rs
```diff
@@ -496,7 +525,109 @@ impl<T: UnderlyingBuffer> ParserIter<'_, T> {
...
+
+                    if self.buf.len() < 2 {
+                        return None;
+                    }
+                    if self.buf[1] != SYNC_CHAR_2 {
+                        self.buf.drain(1);
+                        continue;
+                    }
+
+                    if self.buf.len() < 6 {
```
CramBL, line 597: Why 6? (magic number)

CramBL (commented):
There's so many magic numbers, ideally they are all resolved before the merge, I just commented some of the most egregious ones.

ublox/src/parser.rs
gwbres: I totally agree, but considering the workload and that it is out of scope of the proposal, it's fine if we deal with those in separate PR (these are in the UbxParser after all, which already exists)

gwbres: #126

ublox/src/parser.rs
andrei-ng: Hahaha. Then we should switch it :grin:

andrei-ng:
> I don't think this belongs to this library, for the simple reason that it is already complex enough, and people interested in parsing RTCM from their ublox have another possible scenario (same applies to NMEA). They can simply activate that stream using the UBX library, and switch over to RTCM listening. And the dual case RTCM/UBX can be handled externally. Is there any advantage this proposal would have over handling everything externally ?

Good point @gwbres. We should do one thing and one thing well, not many and poorly. 

I can only think of one use case, where you want to have access multiple streams from within the same application from the same serial connection, e.g., UBX and NMEA. Currently, the UBX parser is globbing everything and nothing escapes this lib's parser except the UBX messages that get decoded. If one wants to have access to UBX, NMEA and/or RTCM from the same serial, using our current version of the library ... it is not possible. 

IMO something along the lines of the current PR is useful as it opens our library for more applications and we can leave it up to our users to decode the RTCM packages as they need as long as we make the raw data available to them.

E.g., With this crate I faced a situation where I wanted to read the NMEA GGA message from the same UBX device in order to send it back to the NTRIP caster which was providing the RTCM corrections. With the current implementation, as our lib's parser is eating everthing except I had to reconstruct the GGA message from multiple UBX messages.

Anyway, it is up for discussion here with you and @CramBL  if we want to have this in or drop it completely.

gwbres:
> If one wants to have access to UBX, NMEA and/or RTCM from the same serial, using our current version of the library ... it is not possible.
> E.g., With this crate I faced a situation where I wanted to read the NMEA GGA message from the same UBX device in order to send it back to the NTRIP caster which was providing the RTCM corrections. With the current implementation, as our lib's parser is eating everthing except I had to reconstruct the GGA message from multiple UBX messages.

Good enough for me. It won't cost much to host just one more struct (UbxRtcm) at this point, if we stick it to just that (=exposing the raw bytes and leaving all interpretation outside)

> Good point @gwbres. We should do one thing and one thing well, not many and poorly.

always better to think first and avoid iterating forever IMO ;)

CramBL:
> > If one wants to have access to UBX, NMEA and/or RTCM from the same serial, using our current version of the library ... it is not possible.
> > E.g., With this crate I faced a situation where I wanted to read the NMEA GGA message from the same UBX device in order to send it back to the NTRIP caster which was providing the RTCM corrections. With the current implementation, as our lib's parser is eating everthing except I had to reconstruct the GGA message from multiple UBX messages.
> 
> Good enough for me. It won't cost much to host just one more struct (UbxRtcm) at this point, if we stick it to just that (=exposing the raw bytes and leaving all interpretation outside)
> 
> > Good point @gwbres. We should do one thing and one thing well, not many and poorly.
> 
> always better to think first and avoid iterating forever IMO ;)

Yep, let's have this to support the use-case mentioned by @andrei-ng 👍

CramBL (approved):

gwbres (approved):

andrei-ng:
@cirdeirf , if you have time to submit a PR with that addition, it would be helpful. We just didn't get the time to do it.

gwbres:
> I do not necessarily propose to parse RTCM and NMEA messages. This (and my) PR do not intend to do so but just expose the possibility to not "loose" the data when using this lib's parser. 

That's what we did for RTCM and is the best thing to do, it is up to the end user to deploy whatever parser they want, we only preserve the data. 

> By the way: Thank you for your fast replies (and this lib). I hope, I do not annoy you with my questions/comments

our pleasure, you're not annoying anyone and any contribution is always welcome

PR #128, opened 2025-04-24
Introduce UbxConstants structure (private)

### LINKED ISSUE #126: Define more constants and replace magic numbers
The UbxParser uses quite a few hardcoded values, and it would be nice to replace them by meaningful constants

gwbres:
- Closes #126

### DIFF
--- a/ublox/src/ubx_packets.rs
+++ b/ublox/src/ubx_packets.rs
@@ -14,9 +14,19 @@ pub trait UbxPacketMeta {
     const MAX_PAYLOAD_LEN: u16;
 }
 
-pub(crate) const SYNC_CHAR_1: u8 = 0xb5;
-pub(crate) const SYNC_CHAR_2: u8 = 0x62;
-pub(crate) const RTCM_SYNC_CHAR: u8 = 0xd3;
+/// Constants used when parsing UBX protocol.
+pub(crate) struct UbxConstants;
+
+impl UbxConstants {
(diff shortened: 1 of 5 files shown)


ublox/src/ubx_packets.rs
```diff
@@ -14,9 +14,28 @@ pub trait UbxPacketMeta {
...
+pub(crate) struct UbxConstants;
+
+impl UbxConstants {
+    /// Size of the UBX header frame markers (in bytes)
+    pub const SYNC_SIZE: usize = 2;
+
+    /// Size of the Payload length encoding (in bytes)
+    pub const PAYLOAD_ENCODING_SIZE: usize = 2;
+
+    /// Total size of the UBX header (SYNC +CLASS +ID +LENGTH)
```
andrei-ng, line 27: Should we add a `pub const CLASS_ID_SIZE: 2`
I see this already pops up in two places, here and in `parser.rs`

```
619: // Payload length identification
           let offset = UbxConstants::SYNC_SIZE + 2; // CLASS + ID
```

andrei-ng (approved):

ublox/src/ubx_packets.rs
gwbres: I will get back to this tomorrow

CramBL: Should it be `pub` or `pub(crate)`?

Issue #121, opened 2025-04-22
Make payload lifetime specifier a macro argument


andrei-ng:
As highlighted in the discussion in https://github.com/ublox-rs/ublox/pull/106#discussion_r2050621827, the specifer `'a` is hardcoded in the macro and cannot be changed or when changed the compiler errors message is quite baffling.

PR #130, opened 2025-04-27
refactor macros into more files and start handling generics

CramBL:
Cherry-picked my commits that touch the derive crate from https://github.com/ublox-rs/ublox/pull/124

resolves https://github.com/ublox-rs/ublox/issues/121

Adds some macro debug utilities as well.

... and fixes the issue @andrei-ng describes in https://github.com/ublox-rs/ublox/pull/124


andrei-ng:
Hey @CramBL, I will have a close look in the coming days. I was off & away for an extended weekend.

andrei-ng:
Great work! 

It is a big refactoring indeed. I tried to look carefully at the maco changes as I am no expert in that regard and nothing specifically stands out. 

How do you want to proceed? Do you want to merge this in and then see what else needs to be done for the CfgVal and other messages that need `recv_send` ?

andrei-ng (approved):

CramBL:
> How do you want to proceed? Do you want to merge this in and then see what else needs to be done for the CfgVal and other messages that need `recv_send` ?

Yes. We can always make new changes to the macros, and this PR will make them much easier to debug etc.

But the big challenge is that it was never designed with generics in mind, handling lifetimes is just an after thought so we'd have to completely redesign the macro crate if we want to properly handle a bunch of lifetimes.

andrei-ng:
> But the big challenge is that it was never designed with generics in mind, handling lifetimes is just an after thought so we'd have to completely redesign the macro crate if we want to properly handle a bunch of lifetimes.

Understood. I think once we get the crate(s) in a stable state we should do that. Currently we are just catching up w.r.t. "technical debt"  ... :D

PR #133, opened 2025-04-30
simplify features set usage for the examples

andrei-ng:
I've simplified a bit the Cargo.toml's of the examples and removed unecessary duplication of feature specification.


CramBL (approved):
Nice!

PR #137, opened 2025-05-09
RXM-SFRBX: GPS L1 C/A message 1, 2 and 3 interpretation

gwbres:
This PR adds a first solution to `UBX-RXM-SFRBX` interpretation, as required by manual navigation.
This is the orbital equivalent of `RXM-RAWX`.

The decoding is unlocked by a new crate option, although this is fairly small, and still does not require `std`.
The decoding logic is `GNSS` dependent. This only handles GPS L1 C/A messages


gwbres:
@andrei-ng, @CramBL,

still a few errors, decoding (scaling/interpretation actually) is not perfect, I hope to conclude this draft in the next few days.
I will keep you updated and provide my test results when I'm done

ublox/src/error.rs
```diff
@@ -45,7 +45,7 @@ impl fmt::Display for ParserError {
         match self {
             ParserError::InvalidChecksum { expect, got } => write!(
                 f,
-                "Not valid packet's checksum, expect {:x}, got {:x}",
+                "Not valid packet's checksum, expect 0x{:02x}, got 0x{:02x}",
```
gwbres, line 48: emphasize this is hexa encoding/printing

ublox/tests/mod.rs
```diff
@@ -0,0 +1,4 @@
+mod generator_test;
+mod parser_binary_dump_test;
+mod parser_tests;
+mod rxm_sfrbx;
```
gwbres, line 4: I am unable to execute the tests individually without this

gwbres:
Let me know what you guys think with respect how to handle the GNSS protocols.
It seems to me that handling GPS externally (hence, having only the iterator / data words rework and then forward) would be feasible. Not sure this will apply to others

ublox/src/error.rs
CramBL: Do you prefer that way of formatting it? Otherwise `{:#02x}` achieves the same as `0x{:02x}`

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=28034d7258915efabaa04f7964cb117e

CramBL:
> I'm not a fan of seeing this code base grow while it could be handled externally. Let me see if I can provide a solution

You could put it in another crate in the workspace, it's easy to remove it or feature gate it then.

gwbres:
> > I'm not a fan of seeing this code base grow while it could be handled externally. Let me see if I can provide a solution
> 
> You could put it in another crate in the workspace, it's easy to remove it or feature gate it then.

Yep, this will take a while, but is the right thing to do. Future protos (if they differ to much from original specs) will have to be handled internally

ublox/src/error.rs
gwbres: The 0x prefix is important

gwbres:
Ok I made it.

I extracted the GPS /QZSS decoding logic in a separate lib, vastly reducing the code quantity.

Interpretation of the `UBX-RXMSFRBX` frame is GNSS dependent, and gated on a `sfrbx` compilation option.
The two tests still pass.

It has not been easy, because GPS is not aligned to 32 bit , but 30.
And Ublox obviously aligns to 32 bit. And since I extracted everything in a dedicated lib, might as well cover the real GPS use case, coming from an unpadded stream of raw bytes. Although this use cas is not 100% verified yet, the interface allows it and it should be fairly close. Padded and aligned data, as this use case require, does work.

ublox/src/ubx_packets/packets/rxm_sfrbx.rs
```diff
@@ -50,3 +69,62 @@ impl core::iter::Iterator for DwrdIter<'_> {
...
+                // either not supported or not applicable
+                None
+            },
+        }
+    }
+
+    #[cfg(feature = "sfrbx-gps")]
+    /// [GpsQzssFrame] interpretation attempt, from parsed UBX-SFRBX data.
+    fn gps_qzss_decoding(&self) -> Option<GpsQzssFrame> {
+        let mut decoder = GpsQzssDecoder::default().without_parity_verification();
```
gwbres, line 109: The decoder does not allocate anything and is `no-std` compliant.
So It's fine to "build" one per interpretation.
This will apply similarly to other cases (when we unlock them), considering GPS frames are tiny streams of bytes

CramBL (changes requested):
Just a few wishes for documentation basically

ublox/tests/rxm_sfrbx/gps.rs
```diff
@@ -0,0 +1,327 @@
+use ublox::{PacketRef, Parser, RxmSfrbxInterpreted};
+
+use gnss_protos::GpsQzssSubframe;
+
+#[test]
+#[cfg(feature = "ubx_proto23")]
+fn sfrbx_gps_eph1() {
+    let bytes = [
```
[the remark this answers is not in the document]
gwbres: This tests one of the possible frame that may be transported by RXM-SFRBX, GPS L1/CA in this very case, because that is the only frame that I have treated so far. So it tests two things:

* that GPS L1/CA RXM-SFRBX is correctly decoded by Ublox
* that GPS L1/CA inner content is correctly interpreted after Ublox transport

ublox/tests/rxm_sfrbx/gps.rs
```diff
@@ -0,0 +1,327 @@
...
+            Err(e) => panic!("UBX-SFRBX parsing failed with {}", e),
+        }
+    }
+    assert!(test_passed, "UBX-SFRBX test failed");
+}
+
+#[test]
+#[cfg(feature = "ubx_proto23")]
+fn sfrbx_gps_eph2() {
+    let bytes = [
```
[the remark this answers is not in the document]
gwbres: * second frame type that may be transported by RXM-SFRBX on the GPS topic

ublox/tests/rxm_sfrbx/gps.rs
```diff
@@ -0,0 +1,327 @@
+use ublox::{PacketRef, Parser, RxmSfrbxInterpreted};
+
+use gnss_protos::GpsQzssSubframe;
+
+#[test]
+#[cfg(feature = "ubx_proto23")]
+fn sfrbx_gps_eph1() {
```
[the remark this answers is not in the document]
gwbres: correct UBX decoding of UBX-RXM-SFRBX, and correct GPS Eph#1, Eph#2 and Eph#3 transport and interpretation

ublox/tests/rxm_sfrbx/gps.rs
```diff
@@ -0,0 +1,327 @@
...
+            Err(e) => panic!("UBX-SFRBX parsing failed with {}", e),
+        }
+    }
+    assert!(test_passed, "UBX-SFRBX test failed");
+}
+
+#[test]
+#[cfg(feature = "ubx_proto23")]
+fn sfrbx_gps_eph3() {
+    let bytes = [
```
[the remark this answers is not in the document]
gwbres: SFRBX GPS Ephemeris frame n°3 is about as accurate as it can get ? 

I added comments on top of that function. 

I will add similar tests when we introduce other interpretation,
but this made me realize the amount of UBX packet tests is very limited. So it's good that here we're adding RXM-SFRBX packet tests, but that should be done (in the future) for other packets as well (especially, complex packets)

PR #143, opened 2025-05-23
UBX + RTCM improvements

gwbres:
* introduce rtcm crate feature, to attempt RTCM decoding
   * add more comments to existing parsers and structures, in particular that UbxRtcm handles both and will adapt on the fly

### DIFF
--- a/ublox/src/parser/ubx_rtcm.rs
+++ b/ublox/src/parser/ubx_rtcm.rs
@@ -1,0 +1,129 @@
+//! Adaptative UBX + RTCM parser
+
+use crate::{
+    error::ParserError,
+    parser::{ubx::extract_packet_ubx, AnyPacketRef, DualBuffer, NextSync, UnderlyingBuffer},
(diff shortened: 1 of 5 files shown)


ublox/src/parser/ubx_rtcm.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let (consumed, msg_frame) = match maybe_data {
+        Ok(data) => find_next_rtcm_frame(data),
+        Err(e) => {
+            return Some(Err(e));
+        },
+    };
+
+    if consumed > 0 {
+        // remove bytes that have been consumed by RTCM identification attempt
```
gwbres, line 71: If the RTCM decoder did consume something, ideally we should advance within our buffer

gwbres:
I'm actually unable to test this in the real world, because I only have an M8T here, which does not support RTCM (up/down links).

I may add a unit test, it's always a good idea

gwbres:
This will require more work, I'll convert it to draft

PR #144, opened 2025-05-23
UBX-MGA: introduce new packets

gwbres:
* Add UBX-MGA-GAL Ephemeris
    * Add UBX-MGA-BDS Ephemeris

### DIFF
--- a/ublox/src/ubx_packets/packets/mga_bds_eph.rs
+++ b/ublox/src/ubx_packets/packets/mga_bds_eph.rs
@@ -1,0 +1,124 @@
+#[cfg(feature = "serde")]
+use super::SerializeUbxPacketFields;
+#[cfg(feature = "serde")]
+use crate::serde::ser::SerializeMap;
+
+use crate::{error::ParserError, UbxPacketMeta};
+use ublox_derive::ubx_packet_recv;
+
+/// UBX-MGA-BDS Ephemeris packet.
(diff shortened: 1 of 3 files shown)


gwbres:
This PR adds definition of UBX MGA Ephemeris for both Galileo and BDS constellations, which might be used in aiding navigation for both of those. I will most likely modify the UBX MGA GPS that I published years ago, as it coud benefit:

* emphasize in the unit being used, because semicircles as provided by the ublox firmware, are not compatible with calculations directly
* add more comments

ublox/src/ubx_packets/packets/mga_bds_eph.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Reserved
+    reserved1: u8,
+
+    /// Autonomous Satellite Health H1 flag
+    sat_h1: u8,
+
+    /// Issue of Data, Clock
+    iodc: u8,
+
+    /// a2 correction term (in seconds.s⁻²)
```
andrei-ng, line 31: The units for this field in the docstring is incorrect, right? It is different then in the other fields, should we make it consistent ? (seconds / seconds^2) Or something similar?

andrei-ng (approved):

andrei-ng:
Just that comment regarding the doc line. The rest looks all right

CramBL (approved):

PR #158, opened 2025-08-03
Add support for some CFG-NAVSPG-* messages

CramBL:
I'll add the rest in another PR once we agree on the implementation of this.

Reference: https://cdn.sparkfun.com/assets/f/7/4/3/5/PM-15136.pdf


ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -1288,3 +1362,49 @@ pub enum TModePosType {
...
+    /// 3D only
+    Only3D = 2,
+    /// Auto 2D/3D
+    #[default]
+    Auto = 3,
+}
+
+#[derive(Debug, Default, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum NavSpgUtcStandard {
```
gwbres, line 1380: Hello,

no need to introduce, this already exists as `UtcStandardIdentifier` (used in other messages)

ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -1288,3 +1362,49 @@ pub enum TModePosType {
...
+    /// UTC as operated by the former Soviet Union; derived from GLONASS time
+    SU = 6,
+    /// UTC as operated by the National Time Service Center, China; derived from BeiDou time
+    NTSC = 7,
+}
+
+/// Dynamic platform model
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum NavSpgDynamicModel {
```
gwbres, line 1395: same here, already exist and remains compatible to the previous Nav5 message,
we should use `CfgNav5DynModel` (possibly renamed to `DynamicModel` to make it less tied to "CfgNav5")

ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -1288,3 +1362,49 @@ pub enum TModePosType {
     /// Lat/Lon/Height position
     LLH = 1,
 }
+
+#[derive(Debug, Default, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum NavSpgFixMode {
```
gwbres, line 1368: Compatible with `CfgNav5FixMode` which should possibly be renamed and used twice as well

gwbres:
Other than that, I don't have many more comments,

considering the already huge code quantity, I think it would be best to reduce it any time we can,
I don't think that should cause any issues, obviously those values are compatible with previous versions of the protocol

ublox/src/ubx_packets/cfg_val.rs
CramBL: It doesn't seem to match. This is from the docs I link to in the PR 
{68EA27AC-A95F-4B23-BCB3-1002B27255D9}

ublox/src/ubx_packets/cfg_val.rs
CramBL: renamed to `NavFixMode` and added entry in changelog under `Breaking`

andrei-ng (approved):

ublox/src/ubx_packets/cfg_val.rs
CramBL: Ok I see that the bike is from proto v31

CramBL: I renamed it to `NavDynamicModel` and protected the WristWatch and Bike variants under feature flags, and changed the default to be the `Portable` variant as specified in the docs

PR #168, opened 2025-08-30
dev nav pvt

CramBL:
- **divide protocol specific NAV-PVT implementations into modules under the nav_pvt module**
- **ensure generated extended bitflags `from` and `into_raw` functions are visible outside of the module they are generated in**
- **move NavPvtFlags and NavPvtFlags2 into common module for NAV-PVT to eliminate duplicate code**


gwbres (approved):
looks good to me, splitting into separate revision is the natural solution

PR #173, opened 2025-09-08
Fix MGA definitions 

gwbres:
- Alll MGA are both RX and TX
- We don't have a TX only type, so the only option is mark TX+RX (not really important)
- Fixed scaling in existing MGA IONO frames
- One or two existing frames were not declared at top level, fixed that
- introduced a few MGA IONO frames
- introduced a few MGA TIME frames

### DIFF
--- a/ublox/src/ubx_packets/packets/packetref_proto17.rs
+++ b/ublox/src/ubx_packets/packets/packetref_proto17.rs
@@ -33,7 +33,7 @@ define_recv_packets!(
         MgaAck,
         MgaGpsIono,
         MgaGpsEph,
-        MgaGpsTime,
+        MgaGpsUtc,
         MgaGalEph,
         MgaGalTime,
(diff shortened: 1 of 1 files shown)


gwbres:
Hello,

I will try to squeeze other MGA frames in 0.7 if I have time (pending PR and next release), but that is not very important.
It's more important to me to have correct definition of existing frames

gwbres:
The RINEX feature will help validate all those scalings.
These are possibly the most complex case to deal with all those scalings.
There's even an issue or potential error in the docs, where GPS-IONO is supossedly expressed in semicircles, while all other XXX-IONO are supposedly expressed in radians. I think the semicircles does not make sense, we will inquire all of this, but it will take time. I'll just leave it open for your review and for a while, now that I can play with all of this, I might be able to a spot a few issues here & there still

PR #174, opened 2025-09-08
PacketOwned.to_ref()

gwbres:
This patch adds the ability to create a PacketRef from a PacketOwned


CramBL:
hmm, this is not the right approach imo. Better to impl `From<PacketOwned> for PacketRef` and/or `AsRef`.

We should use idiomatic Rust instead of implementing all the core traits as random custom functions

ublox_derive/src/output/gen_recv_code.rs
```diff
@@ -68,12 +69,17 @@ pub fn generate_recv_code_for_packet(dbg_ctx: DebugContext, pack_descr: &PackDes
                 &self.0
             }
 
+            pub fn to_ref(&self) -> #ref_name {
+                self.into()
+            }
+
```
CramBL, line 75: Should be an implementation of [AsRef](https://doc.rust-lang.org/std/convert/trait.AsRef.html) instead

```suggestion
```

gwbres: well that remark could apply to existing structures as well..
I'm able to implement a referecing method like this, but unable to implement AsRef which seems more complex to implement

PR #179, opened 2025-09-14
enable all protocol features

CramBL:
Obviously this massively breaks the API, but I think roughly this is the kind of API we want.

We should still unify PacketRef/PacketOwned in the long term, but this will do it for now I think.

- [x] Fix tests
- [x] Fix examples


gwbres:
@CramBL I agree this is on the right track,
keep pushing for it I will give it a try either tomorrow or early this week, unfortunately I won't have time for that today

ublox/src/parser.rs
```diff
@@ -1,19 +1,172 @@
...
+    Parser::new(Vec::new())
+}
+
+/// Creates a parser for UBX Protocol v31 with a fixed-size buffer.
+///
+/// Requires the `ubx_proto31` feature.
+#[cfg(feature = "ubx_proto31")]
+pub fn proto31_with_buffer(buf: &mut [u8]) -> Parser<FixedLinearBuffer<'_>, Proto31> {
+    Parser::new(FixedLinearBuffer::new(buf))
+}
```
CramBL, line 169: This should probably just be deleted. We can write a parser builder if needed, but it's probably overkill.

examples/basic-cli/src/main.rs
```diff
@@ -43,46 +54,142 @@ fn main() {
...
-                    if has_posvel {
-                        let pos: Position = (&pvt).into();
-                        let vel: Velocity = (&pvt).into();
-                        println!(
-                            "NavPvt: Latitude: {:.5} Longitude: {:.5} Altitude: {:.2} m, Speed: {:.2} m/s Heading: {:.2} degrees",
-                            pos.lat, pos.lon, pos.alt
-                            ,vel.speed, vel.heading
-                        );
-                        println!("NavPvt full: {pvt:?}");
+        device.on_data_available(|packet| {
```
andrei-ng, line 57: Wouldn't it be better to create 3 separate examples? What do you think?

andrei-ng (approved):

examples/basic-cli/src/main.rs
CramBL: Yes. Can we handle it in a following PR? Maybe we can also make a more minimal example. We don't have any simple examples.

PR #182, opened 2025-09-20
divide top level protocol packet implementations into submodules

CramBL:


ublox/src/lib.rs
```diff
@@ -15,6 +15,49 @@ pub use crate::{
...
+/// Most users will only need one protocol, so with only one protocol feature enabled
+/// the enum will contain a single variant which is the UBX packet variant of the selected protocol.
+///
+/// # Example
+///
+/// ```rust,ignore
+/// # use ublox::UbxPacket;
+/// match packet {
+///     #[cfg(feature = "ubx_proto14")]
+///     UbxPacket::Proto17(p) => { /* handle proto17 */ }
```
andrei-ng, line 52: Make this 14. As probably 17 was a mistake

PR #183, opened 2025-09-20
examples

CramBL:
- **divide top level protocol packet implementations into submodules**
- **add simple reading from serial device and parsing example**

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,13 +1,31 @@
 # Examples
 
 This folder contains the following examples
+ - [simple-parse](./simple-parse)
  - [basic-cli](./basic-cli)
  - [send-receive](./send-receive)
  - [ublox-tui](./ublox-tui)
  - [DDS](./dds/)
 
+# simple-parse
 
-# uBlox Device
+Minimal example of opening a serial device, reading and parsing ubx packets from it.
+
+This example requires the device to have been preconfigured to output messages.
+
+To run the example with uBlox protocol 23 run
+
+```shell
+cargo run -p simple-parse -- -p /dev/ttyACM0
+```
+
+To run it for other protocol versions, first disable the default features and then enable the specific features needed, e.g.,
+
+```shell
+cargo run -p simple-parse --no-default-features --features alloc,ubx_proto27 -- -p /dev/ttyACM0 
+```
+
+# uBlox Device - This kind of ruins all the examples because the device is so generic and complex!
 
 The device configuration and the reading of packets as well as some common CLI arguments have been abstracted away into an `ublox-device` library. This library is for convenience only as it is used throughout all of the examples.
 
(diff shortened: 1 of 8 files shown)


examples/send-receive/Cargo.toml
```diff
@@ -10,14 +10,14 @@ version = "0.1.0"
...
+ubx_proto14 = ["ublox-device/ubx_proto14"]
+ubx_proto23 = ["ublox-device/ubx_proto23"]
+ubx_proto27 = ["ublox-device/ubx_proto27"]
+ubx_proto31 = ["ublox-device/ubx_proto31"]
 
 [dependencies]
 chrono = "0.4"
 clap = { version = "4.2", features = ["cargo"] }
-serialport = "4.2"
+serialport = { version = "4.2", default-features = false }
```
andrei-ng, line 21: Was this the leading cause of the `libudev` in CI?

examples/README.md
```diff
@@ -1,13 +1,31 @@
...
+To run the example with uBlox protocol 23 run
+
+```shell
+cargo run -p simple-parse -- -p /dev/ttyACM0
+```
+
+To run it for other protocol versions, first disable the default features and then enable the specific features needed, e.g.,
+
+```shell
+cargo run -p simple-parse --no-default-features --features alloc,ubx_proto27 -- -p /dev/ttyACM0
```
andrei-ng, line 25: I am not sure the `simple-parse` example will work with that command as it has no features, right?

examples/README.md
```diff
@@ -1,13 +1,31 @@
...
+```shell
+cargo run -p simple-parse -- -p /dev/ttyACM0
+```
+
+To run it for other protocol versions, first disable the default features and then enable the specific features needed, e.g.,
+
+```shell
+cargo run -p simple-parse --no-default-features --features alloc,ubx_proto27 -- -p /dev/ttyACM0 
+```
+
```
andrei-ng, line 27: Although I get the point of the comment, I guess we shouldn't merge this comment in, right?

I am fine with remove the `device`. What is the suggestion then, to keep each example in charge of its buffer processing loop ?

andrei-ng (approved):

andrei-ng:
Feel free to merge, provided that you remove the comment `uBlox Device - This kind of ruins all the examples because the device is so generic and complex!` :stuck_out_tongue_winking_eye: . Also let me know what you want to do with the other examples. If you have an idea in mind, I can help with the refactor.

examples/send-receive/Cargo.toml
CramBL: yes

examples/README.md
CramBL: yes I need to fix

andrei-ng:
> > @CramBL Do you want to clean the other examples in this PR as well?
> 
> Clean? What do you mean?

I meant get rid of the ublox-device :D

PR #184, opened 2025-09-21
refactor parser rs

CramBL:
- **divide top level protocol packet implementations into submodules**
- **add simple reading from serial device and parsing example**
- **allow running examples with specific features**
- **disable default features of serialport to avoid depending on udev**
- **remove obsolete libudev installation and avoid persisting credentials in checkout action**
- **only enable std and ubx_proto23 by default and instead introduce the 'full' feature to enable all features**
- **simplify basic-cli example by only handling one protocol at a time**
- **refactor examples, reduce duplication**
- **Buffer and docs improvements**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,7 @@ jobs:
       matrix:
         feature-args:
           - '' # Default features
+          - 'full'
           - --features "alloc std ubx_proto23"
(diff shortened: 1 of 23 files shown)


andrei-ng:
Looks good!

andrei-ng (approved):

PR #185, opened 2025-09-27
make FixedBuffer public

CramBL:


Cargo.toml
```diff
@@ -14,11 +14,13 @@ readme = "README.md"
...
 
 [workspace.lints.clippy]
-manual_assert = "warn"
+# Should we rewrite these?
+# manual_assert = "warn"
 manual_clamp = "warn"
 manual_instant_elapsed = "warn"
 manual_is_power_of_two = "warn"
-manual_let_else = "warn"
+# Consider enabling this if/when we bump the MSRV
```
andrei-ng, line 22: Is something us blocking from bumping MSRV?

PR #186, opened 2025-09-27
parser simplification

CramBL:
- **use wrapping_add instead of overflowing_add where the indicator is unused**
- **Constrain MAX_PAYLOAD_LEN, refine parser checksum implementation**


gwbres (approved):

gwbres:
Thank you for writing more tests, 
I need to take a look at how many tests we have in this lib, but I don't think we have enough.
Maybe at some point we should move the test packet generator you wrote here to we can use it everwhere

PR #187, opened 2025-09-27
Add fuzzing (proptest) for UBX Nav PVT

CramBL:
It's a feature that the test code generation is completely decoupled from our packet definitions that the parser uses


gwbres (approved):
This is very cool thank you very much ! 
I hope to add similar stuff to complex frames like RAWX

andrei-ng (approved):

PR #201, opened 2025-09-27
add mon rf

CramBL:
Builds on top of #191 

Resolves #196

PR #202, opened 2025-09-28
Prepare for new release

gwbres:


andrei-ng:
> @CramBL,
> 
> You have provided a significant amount of new stuff, we should mark this as a new version. Let me know if there's any pending topic you want to combine to this (possibly in other PR), this can remain open

@gwbres , I want to squeeze in one more PR in this new release. I will work on it tonight

gwbres:
> Yea I would also want https://github.com/ublox-rs/ublox/pull/201 included. And btw we never released 0.7.0 so there's no reason to go to 0.8.0.

oh that's very true let me correct that

> @gwbres , I want to squeeze in one more PR in this new release. I will work on it tonight

sound good, no hurries

gwbres:
> > @CramBL,
> > You have provided a significant amount of new stuff, we should mark this as a new version. Let me know if there's any pending topic you want to combine to this (possibly in other PR), this can remain open
> 
> Yea I would also want #201 included. And btw we never released 0.7.0 so there's no reason to go to 0.8.0.

hum looks like we already incremented the version you're right,
let's forget about that branch, all we need is to push a tag at some point

PR #205, opened 2025-09-29
Make ESF-MEAS recv + send

### LINKED ISSUE #123: Make `#[ubx_packet_recv_send]` work with variable length payload

andrei-ng:
- change the send macro implementation to allow an extra field after a repetable block.
 - implement the same pattern as in CfgGnss to be able to construct an EsfMeas packet with the Builder pattern from a slice of Measurement "blocks".

Closes #123

### DIFF
--- a/examples/ublox-tui/Cargo.toml
+++ b/examples/ublox-tui/Cargo.toml
@@ -29,7 +29,11 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 unicode-width = "0.2"
 lazy_static = "1.5"
 tracing-error = "0.2"
-tui-logger = { version = "0.17", features = ["crossterm", "tracing-support"] }
+# can be relaxed to 0.17.4 once we bump MSRV 
+tui-logger = { version = "=0.17.3", features = [
+    "crossterm",
+    "tracing-support",
+] }
 directories = "6.0"
 
 ublox-device = { path = "../ublox-device", default-features = false, optional = true }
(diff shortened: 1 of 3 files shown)


examples/ublox-tui/Cargo.toml
```diff
@@ -3,7 +3,7 @@ authors = ["Andrei Gherghescu <andrei-ng@protonmail.com>"]
 description = "A TUI for visualizing PVT and other UBX messages"
 name = "ublox-tui"
 publish = false
-rust-version.workspace = true
+# rust-version.workspace = true
```
CramBL, line 6: why comment out this?

andrei-ng: It is back. I was fixing tui example, and then I forgot that one in.

andrei-ng:
> Looks good except for a few minor things. I hate to see the macro get more complicated but oh well.

Yes, I know. I was trying not to touch the macro, but unfortunately while I could get away in CfgGnss, here I couldn't. At least not easily. I do hope that we get some time in the future to move to traits.

ublox_derive/src/output/gen_send_code.rs
```diff
@@ -28,21 +29,37 @@ pub fn generate_send_code_for_packet(_dbg_ctx: DebugContext, pack_descr: &PackDe
...
+                        .into_fn;
+                    extend_fields.push(quote! {
+                        {
+                            let buf = #into_fn(self.#name);
+                            let bytes: &[u8] = buf.as_ref();
+                            len_bytes += bytes.len();
+                            out.extend(bytes.iter().copied());
+                        }
+                    });
+                    continue;
```
CramBL, line 61: isn't the `continue` redundant?

ublox_derive/src/output/gen_send_code.rs
```diff
@@ -28,21 +29,37 @@ pub fn generate_send_code_for_packet(_dbg_ctx: DebugContext, pack_descr: &PackDe
...
+                }
```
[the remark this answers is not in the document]
andrei-ng: The order was more in line with how the packet is structured. I agree, that !conditons are not the most readable. 
I changed the order

CramBL (approved):

PR #216, opened 2025-11-02
Derive the Clone trait for owned packets

benfarkas98:


CramBL:
The tests for code generation needs to be updated. It would be better to do these via [insta](https://crates.io/crates/insta) tbh, would be make it clearer and easier to maintain

andrei-ng (approved):

PR #221, opened 2025-11-20
NMEA packet (unparsed) extraction

cirdeirf:
Add support to extract unparsed NMEA packets analogous to RTCM packet extraction (PR #125).
Since parsing has been split to UBX-only and UBX+RTCM, I did the same and added UBX+RTCM+NMEA. Thus keeping it backwards compatible. Though I personally think one parser for all would suffice and users could either configure the device to send only UBX packets or just ignore any returned RTCM/NMEA packets. Splitting it up currently results in unnecessary code duplication.

### DIFF
--- a/ublox/src/parser.rs
+++ b/ublox/src/parser.rs
@@ -3,7 +3,9 @@ use alloc::vec::Vec;
 
 use crate::{
     error::ParserError,
-    ubx_packets::{RTCM_SYNC_CHAR, SYNC_CHAR_1, SYNC_CHAR_2},
+    ubx_packets::{
+        NMEA_END_CHAR_1, NMEA_END_CHAR_2, NMEA_SYNC_CHAR, RTCM_SYNC_CHAR, SYNC_CHAR_1, SYNC_CHAR_2,
+    },
     UbxPacket, UbxProtocol,
 };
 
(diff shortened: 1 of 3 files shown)


ublox/src/parser.rs
```diff
@@ -237,19 +239,42 @@ impl<T: UnderlyingBuffer, P: UbxProtocol> Parser<T, P> {
             _phantom: PhantomData,
         }
     }
+
+    /// Appends `new_data` to the internal buffer and returns and iterator over the buffer
+    /// that will yield [UbxPackets, RtcmPackets, or NmeaPackets](AnyPacketRef) on demand.
+    pub fn consume_ubx_rtcm_nmea<'a>(
```
gwbres: I think we just want to delete the previous `ubx_rtcm` and simply have a single method here, this very one and only.
Don't you guys agree ? @andrei-ng

andrei-ng:
> Sorry, please don't merge yet: Draining the additional NMEA chars at the beginning is not correct. I added that in an attempt to fix the parser_oom_clears_buffer tests where an OutOfMemory Error on a UBX packet left bytes in the buffer which happen to include the NMEA sync char.

Sure. We'll wait. No rush

cirdeirf:
OK, it looks good to me now. Thank you for your patience.

ublox/src/parser.rs
CramBL: I assume we can also avoid any API breakage then

andrei-ng: > I assume we can also avoid any API breakage then

I was thinking the same...

cirdeirf:
Consolidating the consume functions has been reverted and all commits have been squashed into one. As far as I'm concerned it's ready for merging.

andrei-ng (approved):

PR #230, opened 2025-12-11
CFG-MSGOUT-UBX_ESF*, CFG-SF*, and dynamic models

cirdeirf:
Hi,
this adds support for some config keys/values (CFG-MSGOUT-UBX_ESF*, CFG-SF*) and dynamic models (mower, e-scooter, rail).

### DIFF
--- a/ublox/src/ubx_packets/cfg_val.rs
+++ b/ublox/src/ubx_packets/cfg_val.rs
@@ -209,6 +209,12 @@ macro_rules! from_cfg_v_bytes {
             9 => NavDynamicModel::WristWornWatch,
             #[cfg(feature = "ubx_proto31")]
             10 => NavDynamicModel::Bike,
+            #[cfg(feature = "ubx_proto31")]
+            11 => NavDynamicModel::Mower,
+            #[cfg(feature = "ubx_proto31")]
+            12 => NavDynamicModel::EScooter,
+            #[cfg(feature = "ubx_proto31")]
+            13 => NavDynamicModel::Rail,
             _ => unreachable!(
                 "CFG-NAVSPG-DYNMODEL_TYPE value not supported by protocol specification"
             ),
--- a/ublox/src/ubx_packets/packets/cfg_nav5.rs
+++ b/ublox/src/ubx_packets/packets/cfg_nav5.rs
@@ -135,6 +135,12 @@ pub enum NavDynamicModel {
     #[cfg(feature = "ubx_proto31")]
     /// supported in protocol versions 19.2
     Bike = 10,
+    #[cfg(feature = "ubx_proto31")]
+    Mower = 11,
+    #[cfg(feature = "ubx_proto31")]
+    EScooter = 12,
+    #[cfg(feature = "ubx_proto31")]
+    Rail = 13,
 }
 
 /// Position Fixing Mode


ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -209,6 +209,12 @@ macro_rules! from_cfg_v_bytes {
             9 => NavDynamicModel::WristWornWatch,
             #[cfg(feature = "ubx_proto31")]
             10 => NavDynamicModel::Bike,
+            #[cfg(feature = "ubx_proto31")]
+            11 => NavDynamicModel::Mower,
+            #[cfg(feature = "ubx_proto31")]
+            12 => NavDynamicModel::EScooter,
+            #[cfg(feature = "ubx_proto31")]
+            13 => NavDynamicModel::Rail,
```
CramBL: Could you link to where you found these models? I don't see them in the version 31.12 documentation for the ZED-F9H: https://content.u-blox.com/sites/default/files/u-blox_ZED-F9H_InterfaceDescription_%28UBX-19030118%29.pdf

cirdeirf: Sorry, I forgot that this may not be correct. My information is based on the ZED-F9R interface descriptions. The first version where these dynamic models may be found is
https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPS-1.30_InterfaceDescription_UBX-22010984.pdf
which has protocol version 33.30. Though the previous version (https://content.u-blox.com/sites/default/files/F9-HPS120_Interfacedescription_UBX-19056845.pdf) which doesn't support mower, etc.,  has protocol version 33.21.
I did not find any ZED-F9R documentation with with version < 33.

I'll probably have to introduce a new ubx_proto33-feature for this to be correct, won't I?

andrei-ng: > which has protocol version 33.30. Though the previous version (https://content.u-blox.com/sites/default/files/F9-HPS120_Interfacedescription_UBX-19056845.pdf) which doesn't support mower, etc., has protocol version 33.21.
> I did not find any ZED-F9R documentation with with version < 33.

Indeed, I also checked the ICD I have for F9R and these are defined starting with at least 33.20

cirdeirf: OK, I'll look into a ubx_proto33-feature (might be a new pull request).
Is EsfStatus (-Alg, etc.) a proto33-feature? It's already part of this lib (though not as a CfgKey/CfgVal) but I cannot find any mention of ESF in you 31.12 ICD.

andrei-ng: In this particular case it is also related to application /hardware. F9H is for heading applications and does not contain an IMU, hence I suspect that some modes don't make sense if you don't do ESF (external sensor fusion) with external data instead of on board IMU. That would be my guess..

CramBL: > EDIT: the problem with these models is that they seem to be both product specific and protocol specific. E.g., F9R (proto 33.2) has some modes while the latest F9 HPS (proto 33.3) (the one you linked) has extra . So not sure how to approach this. @CramBL , @gwbres any ideas?

It would be great if proto 33 was just a single spec, but imo there's no other choice than to distinguish between 33.2 and 33.3.

I think the best approach is to introduce a `proto33` feature that is the common denominator between all 33+ protocols, and then guard 33.2 specific features with an additional `proto33.2` feature, and similar for 33.3 features.

Then the`proto33.2` and `proto33.3` features should depend on the `proto33` feature.

cirdeirf: Maybe I don't need a proto33 for the dynamic models after all (at least for mower and e-scooter):
This ICD (proto 27.50) from the ZED-F9P (https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPG-1.51_InterfaceDescription_UBXDOC-963802114-13124.pdf) contains both of these (and to be honest I'd be happy with the mower dynamic model and the additional cfg keys/values.

cirdeirf: Out of curiosity: What is the rationale behind the proto-features?
Since the protocol seems to be accumulative so far (no clashing fields/options), one could leave it to the user to only use those options/access those fields that the model in use allows/offers. Of course I can see, that it'd be better if these aren't even offered in the ublox lib (as it is now) but it comes at the cost of manually sighting past ICD (and even then missing some things).

cirdeirf: I've had a look at some ZED-F9* modules. It kind of looks like proto 31 doesn't have the newer models because that ICD/protocol has not been updated since 2020.

|  | BIKE | MOWER | ESCOOTER | RAIL |
|--------|--------|--------|--------|--------|
| [proto 27.50](https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPG-1.51_InterfaceDescription_UBXDOC-963802114-13124.pdf) (F9P, 19-Nov-2024) | x | x | x |  |
| [proto 29.25](https://content.u-blox.com/sites/default/files/documents/u-blox-F9-TIM-2.25_InterfaceDescription_UBXDOC-963802114-13231.pdf) (F9T, 28-Aug-2025) | x | x | x | x |
| [proto 30.50](https://content.u-blox.com/sites/default/files/documents/u-blox-F9-LAP-1.50_InterfaceDescription_UBXDOC-963802114-13052.pdf) (F9K, 13-Mar-2025) | x | x | x | x |
| [proto 31.12](https://content.u-blox.com/sites/default/files/u-blox_ZED-F9H_InterfaceDescription_%28UBX-19030118%29.pdf) (F9H, 19-Nov-2020) |  |  |  |  |
| [proto 33.40](https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPS-1.40_InterfaceDescription_UBXDOC-963802114-13138.pdf) (F9R, 6-Apr-2025) | x | x | x | x | 

Is there another proto 31 ICD that I'm not aware of or why is BIKE part of proto 31 in the lib?
https://github.com/ublox-rs/ublox/blob/2d02da8b5de4f8becfd74b5b902583668bae60ff/ublox/src/ubx_packets/packets/cfg_nav5.rs#L135-L137

cirdeirf:
This pull request has been superseded by #255 which added CFG-MSGOUT-UBX_ESF*, CFG-SF*, and the newest dynamic models for protocol version 33.

PR #231, opened 2025-12-11
Implement From<CfgVal> for CfgKey and add derives for CfgVal/CfgKey

cirdeirf:
Hi,
this implements `From<CfgVal>` and `From<&CfgVal>` for CfgKey using a new `key()` getter and adds some derives for CfgKey (PartialEq, Eq, Hash) and CfgVal (PartialEq).

### DIFF
--- a/ublox/src/ubx_packets/cfg_val.rs
+++ b/ublox/src/ubx_packets/cfg_val.rs
@@ -380,6 +380,7 @@ macro_rules! cfg_val {
           self.len() == 0
       }
 
+      /// Returns the [configuration key](CfgKey) corresponding to this value.
       #[inline]
       pub const fn key(&self) -> CfgKey {
         match self {


ublox/src/ubx_packets/cfg_val.rs
```diff
@@ -380,6 +380,15 @@ macro_rules! cfg_val {
           self.len() == 0
       }
 
+      #[inline]
+      pub const fn key(&self) -> CfgKey {
```
CramBL: I know not everything is equally well documented, but we're trying to improve it gradually :smile: so could you please add a doc comment, however trivial it might be.

CramBL (approved):
Looks good to me, just one doc comment would be nice

andrei-ng (approved):

CramBL:
Thanks @cirdeirf !

Issue #228, opened 2025-12-10
Interest in contributing 6 safety-related UBX messages


gsokoll:
I am looking at developing a ROS2 driver in Rust for ublox ZED-F9P devices for safety-focused industrial automation, and would like to use ublox-rs.  I have identified 6 (currently unsupported) UBX messages that would be valuable for GNSS integrity monitoring.

### Messages I'd like to contribute

| Message | Class | ID | Length | Purpose |
|---------|-------|----|---------| --------|
| NAV-COV | 0x01 | 0x36 | 64 (fixed) | Full position/velocity covariance matrix |
| NAV-POSECEF | 0x01 | 0x01 | 20 (fixed) | ECEF position |
| RXM-COR | 0x02 | 0x34 | 12 (fixed) | Differential correction status (SPARTN/RTK) |
| MON-COMMS | 0x0A | 0x36 | Variable | Communication port diagnostics (relates to #192) |
| SEC-SIG | 0x27 | 0x09 | 16 (fixed) | Jamming/spoofing detection |
| SEC-SIGLOG | 0x27 | 0x10 | Variable | Security event log |

### Use case

These messages enable proper integrity monitoring for autonomous systems:

- **NAV-COV** provides full covariance for sensor fusion (proper EKF weighting rather than scalar accuracy)
- **SEC-SIG / SEC-SIGLOG** provide jamming and spoofing detection status and event history
- **RXM-COR** indicates RTK/PPP/SPARTN correction health
- **MON-COMMS** helps diagnose data flow issues

### Question

I'm happy to contribute PRs for these, but was wondering how you would prefer them to be submitted:
1. **One combined PR** with all 6 messages
2. **Individual PRs** (one per message)
3. **Grouped PRs** by theme:
   - PR 1: NAV-COV, NAV-POSECEF (NAV class additions)
   - PR 2: SEC-SIG, SEC-SIGLOG (new SEC class)
   - PR 3: RXM-COR, MON-COMMS (monitoring/status)

Let me know what works best for the project!  Thanks for maintaining this excellent crate.

andrei-ng:
Hi, @gsokoll . 
I propose option 3, so we keep the PRs size somewhere in the middle, not too big , not too small.

If you could also add proptests, like @CramBL has done in the latest additions, https://github.com/ublox-rs/ublox/pull/189, that would be great.

gsokoll:
[...]

PR #233, opened 2025-12-14
feat: add RXM-COR and MON-COMMS packets

gsokoll:
This PR adds support for the following UBX messages:
- **UBX-RXM-COR** (0x02 0x34): Differential correction input status.
- **UBX-MON-COMMS** (0x0A 0x36): Communication port status and IO statistics.

## Changes
- Implemented `MonComms` with support for variable-length port blocks using a custom iterator (`MonCommsPortIter`).
- Implemented `RxmCor` with bitfield mapping for status info (protocol, error status, message used).
- Registered messages in `packetref_proto27.rs` and `packetref_proto31.rs`.
- Added fuzz tests for both messages.

Partially addresses #228.

### DIFF
--- a/ublox/src/ubx_packets/packets/mon_comms.rs
+++ b/ublox/src/ubx_packets/packets/mon_comms.rs
@@ -118,7 +118,10 @@ impl<'a> MonCommsPortIter<'a> {
     }
 
     /// Validate raw repeated-group payload: must be a multiple of 40 bytes.
-    #[allow(dead_code, reason = "Used by ubx_packet_recv macro for validation, but may appear unused in some feature configurations")]
+    #[allow(
+        dead_code,
+        reason = "Used by ubx_packet_recv macro for validation, but may appear unused in some feature configurations"
+    )]
     fn is_valid(payload: &[u8]) -> bool {
         payload.len() % 40 == 0
     }


ublox/tests/fuzz_rxm_cor.rs
```diff
@@ -0,0 +1,143 @@
...
+        frame_core.push(class_id);
+        frame_core.push(message_id);
+        frame_core.write_u16::<LittleEndian>(length).unwrap();
+        frame_core.extend_from_slice(&payload);
+
+        let (ck_a, ck_b) = calculate_checksum(&frame_core);
+
+        let mut final_frame = Vec::with_capacity(8 + payload.len());
+        final_frame.push(0xB5);
+        final_frame.push(0x62);
```
CramBL: Would be great to have named variables here instead of magic numbers

ublox/tests/fuzz_mon_comms.rs
```diff
@@ -0,0 +1,276 @@
...
+        frame_core.push(class_id);
+        frame_core.push(message_id);
+        frame_core.write_u16::<LittleEndian>(length).unwrap();
+        frame_core.extend_from_slice(&payload);
+
+        let (ck_a, ck_b) = calculate_checksum(&frame_core);
+
+        let mut final_frame = Vec::with_capacity(8 + payload.len());
+        final_frame.push(0xB5);
+        final_frame.push(0x62);
```
CramBL: Can these be named constants?

ublox/Cargo.toml
```diff
@@ -4,7 +4,7 @@ description = "A crate to communicate with u-blox GPS devices using the UBX prot
 name = "ublox"
 homepage = "https://github.com/ublox-rs"
 repository = "https://github.com/ublox-rs/ublox"
-version = "0.8.0"
+version = "0.9.0"
```
CramBL: We will bump the version in another PR when we cut a new release

```suggestion
version = "0.8.0"
```

CramBL (changes requested):
Great work! :star: 

I didn't double check everything with the docs, I assume that you've done your homework, so I just focused on code quality.

There's a couple of very minor things, otherwise it looks great!

ublox/Cargo.toml
gsokoll: This was to address the semver CI jobs that were failing because adding new variants to the public PacketRef enum is considered a breaking change.

I bumped it to 0.9.0 just to get that CI check to pass. Is there a way to suppress or ignore the SemVer check for this PR, or should I leave the version at 0.8.0 and let that specific check fail?

ublox/tests/fuzz_rxm_cor.rs
gsokoll: Have added named constants in the test file

CramBL (approved):
Great! It just seems like you need to run `cargo fmt`, otherwise LGTM

PR #234, opened 2025-12-14
feat: add SEC-SIG and SEC-SIGLOG packets

gsokoll:
This PR adds support for the following security-related UBX messages:
- **UBX-SEC-SIG** (0x27 0x09): Signal security status (jamming/spoofing detection).
- **UBX-SEC-SIGLOG** (0x27 0x10): Signal security event log.

## Changes
- Implemented `SecSig` with support for variable-length jamming frequency blocks.
- Implemented `SecSiglog` with support for the event log history iterator.
- Added strong typing for security flags (JammingState, SpoofingState).
- Registered messages in `packetref_proto27.rs` and `packetref_proto31.rs`.
- Added fuzz tests.

Partially addresses #228.

### DIFF
--- a/ublox/src/ubx_packets/packets/sec_sig.rs
+++ b/ublox/src/ubx_packets/packets/sec_sig.rs
@@ -62,13 +62,13 @@ impl From<u32> for SecSigJamStateCentFreq {
 
 /// Iterator for SEC-SIG jam state entries.
 #[derive(Debug, Clone)]
-pub struct SecSigJamStateCentFreqIter<'a> {
-    data: &'a [u8],
+pub struct SecSigJamStateCentFreqIter<'d> {
+    data: &'d [u8],
     offset: usize,
 }
 
-impl<'a> SecSigJamStateCentFreqIter<'a> {
-    fn new(data: &'a [u8]) -> Self {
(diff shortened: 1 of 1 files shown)


ublox/src/ubx_packets/packets/sec_sig.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Debug, Clone)]
+pub struct SecSigJamStateCentFreqIter<'a> {
+    data: &'a [u8],
+    offset: usize,
+}
+
+impl<'a> SecSigJamStateCentFreqIter<'a> {
+    fn new(data: &'a [u8]) -> Self {
+        Self { data, offset: 0 }
+    }
```
CramBL: Tiny thing, but giving the lifetime a slightly less generic name would be nice.

I just suggest `d` for `data`.

```suggestion
pub struct SecSigJamStateCentFreqIter<'d> {
    data: &'d [u8],
    offset: usize,
}

impl<'d> SecSigJamStateCentFreqIter<'d> {
    fn new(data: &'d [u8]) -> Self {
        Self { data, offset: 0 }
    }
```

CramBL (approved):
looks great!

PR #246, opened 2026-01-03
bump msrv to 1.85

### LINKED ISSUE #245: CI is failing in build_examples (dds) and msrv. The dds example's dependency async-lock@3.4.2 now requires rustc 1.85, but CI uses 1.83. Does upstream need to update MSRV or pin dependencies?
CI is failing in build_examples (dds) and msrv. The dds example's dependency async-lock@3.4.2 now requires rustc 1.85, but CI uses 1.83. Does upstream need to update MSRV or pin dependencies?

_Originally posted by @gsokoll in https://github.com/ublox-rs/ublox/issues/240#issuecomment-3681188134_

andrei-ng:
Closes #245

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install MSRV
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.83.0
+          toolchain: 1.85.0
           override: true
           components: rustfmt, clippy
       - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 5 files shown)


gwbres:
Yeah I agree with that statement 100%.
It's normal that an application demands more than the core library, but we should find someway that it does not reflect on the core library itself (the simpler the better when it comes to a library). How do we do this then? the examples are already split in the CI/CD, we can simply increment the MSRV of that section

andrei-ng:
Closed in favor of #247

PR #247, opened 2026-01-04
Bump msrv examples

### LINKED ISSUE #245: CI is failing in build_examples (dds) and msrv. The dds example's dependency async-lock@3.4.2 now requires rustc 1.85, but CI uses 1.83. Does upstream need to update MSRV or pin dependencies?

andrei-ng:
Alternative for #246 

Closes #245

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Code style
         run: |
           cargo fmt --all -- --check
-          cargo clippy --all-features --all-targets -p ublox -p ublox_derive -- -D warnings
+          cargo clippy --all-features --all-targets -- -D warnings
 
 
   build:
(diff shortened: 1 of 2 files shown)


andrei-ng:
Alternative, as disscuseed in https://github.com/ublox-rs/ublox/pull/246

andrei-ng:
> This works and it's fine with me. But ideally we use the Cargo.toml default-members or whatever it's called, so we don't have to specifically include the two main crates, and instead it's just the default mode.
> 
> But it's just a nitpick basically, it's not like we anticipate having more and more crates every month so this works just fine.

Yup. Very good point. Updated

Issue #236, opened 2025-12-15
"lastCorrectionAge" is present in "flags3" (NAV-PVT) for proto 27


CramBL:
Maybe I don't need a proto33 for the dynamic models after all (at least for mower and e-scooter):
This ICD (proto 27.50) from the ZED-F9P (https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPG-1.51_InterfaceDescription_UBXDOC-963802114-13124.pdf) contains both of these (and to be honest I'd be happy with the mower dynamic model and the additional cfg keys/values.

Edit: "lastCorrectionAge" is present here in "flags3" (NAV-PVT) as well. The lib currently assumes it isn't for proto27/31.

_Originally posted by @cirdeirf in https://github.com/ublox-rs/ublox/pull/230#discussion_r2618349186_

CramBL:
@cirdeirf can you clarify what you mean? I see it right here:

https://github.com/ublox-rs/ublox/blob/2d02da8b5de4f8becfd74b5b902583668bae60ff/ublox/src/ubx_packets/packets/nav_pvt/proto27_31.rs#L138-L175

gsokoll:
I might be mistaken, but lastCorrectionAge is present in flags3 for proto27 (ZED-F9P HPG 1.32+), however:

- flags3 is defined as u8 but should be u16 (X2 per spec, offset 78)
- The `reserved1`: [u8; 5] should be `reserved0`: [u8; 4] to match the 2-byte flags3

The lastCorrectionAge bits (1-4) are already being read, but the extraction mask needs fixing. Happy to submit a PR including the implementation for `authTime` and `nmaFixStatus` unless someone else wants to dive in first.

<img width="808" height="740" alt="Image" src="https://github.com/user-attachments/assets/edd0763d-e4a5-4e67-bb14-a2d293a8197f" />

<img width="791" height="608" alt="Image" src="https://github.com/user-attachments/assets/202d53fb-c72a-4538-9dec-c2168ecb2b86" />

gsokoll:
@andrei-ng I'm fine with holding off on a PR until #259 is done.

PR #255, opened 2026-01-05
Add basic support for protocol version 33

cirdeirf:
Hi,
this adds basic support for protocol version 33. For now this is mostly a duplicate of proto31 and intended as a first step for any proto33-specific changes. For now, these changes are limited to new dynamic models and some config entries (CFG-MSGOUT-UBX_ESF*, CFG-SF*) as mentioned in #230.

The incompleteness for NAV-PVT (as mentioned in #236) is true for proto33 as well but has not been addressed yet since incorporating that here would have been overwhelming for me.


gsokoll:
1. Unless the maintainers intend to make a new major release, then just don't rename proto27_31.rs as this breaks the public API. Instead, keep proto27_31.rs and update the cfg gate to include proto33:

```
#[cfg(any(
    feature = "ubx_proto27",
    feature = "ubx_proto31",
    feature = "ubx_proto33",  // Add this
))]
pub mod proto27_31;  // Keep original name - kinda misleading since you've added protver 33 support, but non-breaking

```
2. Add #[non_exhaustive] to NavDynamicModel.  Adding Mower, EScooter, Rail variants to an exhaustive enum is a breaking change.  So in cfg_nav5.rs:

```
#[non_exhaustive]  // Add this
#[derive(Default)]
#[ubx_extend]
#[repr(u8)]
pub enum NavDynamicModel { ... }

```

3. Rebase will require resolving a minor conflict in parser.rs (test ordering near ACK_ACK_BYTES const).

andrei-ng:
> > 2. Add #[non_exhaustive] to NavDynamicModel.  Adding Mower, EScooter, Rail variants to an exhaustive enum is a breaking change.  So in cfg_nav5.rs:
> > 
> > ```
> > #[non_exhaustive]  // Add this
> > #[derive(Default)]
> > #[ubx_extend]
> > #[repr(u8)]
> > pub enum NavDynamicModel { ... }
> > ```
> 
> Marking it `non_exhaustive` seems to be breaking as well or am I misinterpreting the failed semver checks?

You are correct. That is a breaking change as well. 
No worries, we will release a new version after all the changes and all the new packets from @gsokoll .

andrei-ng (approved):

PR #261, opened 2026-01-10
Split NavPvt per proto instead of using an aggregate file for 27 & 31

### LINKED ISSUE #259: Split `proto27_31` into `proto27` and `proto31`

andrei-ng:
Closes #259

With this change, some of the issues raised in #236 can be addressed in a cleaner way.


ublox/src/ubx_packets/types.rs
```diff
@@ -47,6 +46,22 @@ pub struct Velocity {
...
+    fn ground_speed_2d(&self) -> f64;
+    fn heading_motion(&self) -> f64;
+    fn year(&self) -> u16;
+    fn month(&self) -> u8;
+    fn day(&self) -> u8;
+    fn hour(&self) -> u8;
+    fn min(&self) -> u8;
+    fn sec(&self) -> u8;
+    fn nanosec(&self) -> i32;
+}
```
CramBL: I understand the motivation but this is not ideal. A trait based approach will ultimately be the best for the crate and for users, but what this achieves right now, is that all the generated documentation is now hidden behind another layer of indirection, making it useless in practice.

CramBL: Wait I didn't realize the trait isn't public, and it's not re-exported?

In that case it achieves even less, downstream libraries cannot use it to generalize their implementation.

I want to hear your thoughts, but for now I suggest 2 changes:

1. Copy+paste the documentation for each of these functions to the trait, so that the docs are never hidden.
2. Make a declarative macro that can implement the boilerplate glue code for the protocol-specific structs such that we can shave off a lot of lines on this PR.

CramBL: Let's start by just having it an internal trait, so we can better evaluate it, in smaller future updates. The big risk is making it public and then discovering that some function the trait defines, cannot be implemented by one or more of the protocols, or that newer protocols invalidate it. Then it would need to be a `NavPvtFieldsProto33AndEarlier` and would obviously be a breaking change.

Also I'm not satisfied with the trait name at this point, so we can also jam on that a bit.

andrei-ng:
@CramBL , I for one, think the latest commit looks better, thoughts?

PR #290, opened 2026-05-31
fix(mon-comms): decode PortId from the high byte of portId

### LINKED ISSUE #288: Decoding of UBX-MON-COMMS portId field
## Summary

`PortId::from(u16)` in `ublox/src/ubx_packets/packets/mon_comms.rs` treats the UBX-MON-COMMS `portId` field as a 0 to 5 index. On u-blox hardware `portId` is a 16-bit value whose high byte identifies the interface, so only I2C decodes correctly and every other port (UART1, UART2, USB, SPI) falls through to `PortId::Unknown(...)`.

## Current code

from line 64 of `ublox/src/ubx_packets/packets/mon_comms.rs`

```rust
impl From<u16> for PortId {
    fn from(value: u16) -> Self {
        match value {
            0 => PortId::I2c,
            1 => PortId::Uart1,
            2 => PortId::Uart2,
            3 => PortId::Usb,
            5 => PortId::Spi,
            other => PortId::Unknown(other),
        }
    }
}
```

The 0 to 5 values currently used appear to mostly align with the separate `txErrors.outputPort` sub-field (`bits 4..2`: `0=N/A,1=I2C,2=UART1,3=UART2,4=USB, 5=SPI`), which is a different field with a different encoding.

## Documented portId

The port numbers reported in MON-COMMS are documented in the u-blox integration manuals under "Communication ports" / "Port number assignment".  It is described as a `U2` "Unique identifier for the port".

The port numbers from the integration manuals are shown below.  The high byte selects the interface (`0=I2C, 1=UART1, 2=UART2, 3=USB, 4=SPI`), and the low byte appears to be a sub-index that varies by product (UART2 `0x0201` on F9P vs `0x0200` on X20P).

| Interface | `portId` |
|---|---|
| I2C | `0x0000` |
| UART1 | `0x0100` |
| UART2 | `0x0200` (ZED-X20P) / `0x0201` (ZED-F9P) |
| USB | `0x0300` |
| SPI | `0x0400` |

## Impact

Any consumer reading per-port stats from MON-COMMS sees `PortId::Unknown(0x0100)` etc. for UART1/UART2/USB/SPI and cannot identify the port; only I2C resolves.

## Proposed fix

Decode on the high byte (handles the UART2 low-byte variance and is forward compatible), keeping `Unknown(value)` for unrecognized interfaces. I can raise a matching PR.

gsokoll:
# PR: fix(mon-comms): decode PortId from the high byte of portId

Fixes #288.

`PortId::from(u16)` decoded the UBX-MON-COMMS `portId` as a 0–5 index, so only
I2C decoded and every other port returned `Unknown(...)`. Decode on the high
byte instead (`value >> 8`), per the integration manuals; see #288 for full
details.

### DIFF
--- a/ublox/src/ubx_packets/packets/mon_comms.rs
+++ b/ublox/src/ubx_packets/packets/mon_comms.rs
@@ -63,13 +63,22 @@ pub enum PortId {
 
 impl From<u16> for PortId {
     fn from(value: u16) -> Self {
-        match value {
-            0 => PortId::I2c,
-            1 => PortId::Uart1,
-            2 => PortId::Uart2,
-            3 => PortId::Usb,
-            5 => PortId::Spi,
-            other => PortId::Unknown(other),
+        // UBX-MON-COMMS reports `portId` as a 16-bit value whose high byte
+        // selects the physical interface; the low byte is a sub-index that
+        // can vary by product (e.g. UART2 is 0x0201 on ZED-F9P but 0x0200 on
+        // ZED-X20P), so decode on the high byte.
+        //
+        // Refs: ZED-F9P Integration Manual (UBX-18010802) Table 27 "Port number
+        // assignment"; ZED-X20P Integration Manual Table 35; NEO-M9N Integration
+        // Manual (UBX-19014286) Table 13. The values are NOT a plain 0 to 5 index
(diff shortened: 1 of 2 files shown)


andrei-ng (approved):

andrei-ng:
Thank you @gsokoll

Issue #188, opened 2025-09-27
Add fuzzing for all packet variants


CramBL:
Fuzzing NAV PVT surfaced some issues, we should slowly grind us through fuzzing every single variant.

I think we should basically just copy paste the approach for NAV PVT for all variants. It will take time, but I think we'll find plenty of issues, and it will be issues that are very very hard to pin down in the wild.

PR #291, opened 2026-06-02
Extract shared UBX fuzz-test helpers into tests/common

gsokoll:
Per the discussion on #188, this extracts the duplicated boilerplate from the
existing proptest fuzz tests into a shared module, so adding fuzz coverage
for more packets is cheap without generating any verification code from 
the design code.

## Changes

- **New `ublox/tests/common/mod.rs`** with the protocol-level primitives that
  were copy-pasted across every fuzz file:
  - `calculate_checksum()`  Fletcher-16 over class/id/len/payload
  - `build_ubx_frame(class, id, payload)`  full sync + header + payload + checksum
  - `finite_f32()`  proptest strategy excluding NaN/inf
- **All 19 `fuzz_*.rs`** now call `build_ubx_frame()` instead of hand-rolling
  the frame, and drop their local `calculate_checksum()` (and `finite_f32()`
  in `fuzz_nav_cov.rs`).
- **`ublox/tests/lib.rs`**  removed 5 redundant `mod fuzz_*;` declarations.
  Each `tests/*.rs` is already its own auto-discovered test binary, so those
  were double-compiled, and the redeclaration would collide with the new
  `mod common;` path resolution.

Net **-451 lines** (+166 / -617).

## Keeping tests auditable

To address the concern about making it hard to accidentally weaken a suite,
the shared module is deliberately limited to obviously-correct, protocol-level
primitives. Per-packet payload structs and their `*_payload_strategy()` /
`to_bytes()` helpers stay in their own files, so each packet's correctness
remains independently reviewable.

## Verification

- `cargo test -p ublox --features full`  all 25 test binaries pass. Every
  generated frame is fed through the real parser (which validates the
  checksum/envelope), so this confirms `build_ubx_frame` is byte-identical to
  the old inline logic.
- `cargo clippy -p ublox --features full --tests` is clean.

## Notes

Opened as a draft. This is the enabling refactor; coverage is currently 19 of
58 `#[ubx_packet_recv]` packets, and the remaining ~39 become cheap follow-ups
on top of `build_ubx_frame`. Happy to adjust the module's shape or split this
differently if you'd prefer.

### DIFF
--- a/ublox/tests/lib.rs
+++ b/ublox/tests/lib.rs
@@ -3,9 +3,4 @@ mod parser_binary_dump_test;
 mod parser_tests;
 mod rxm_sfrbx;
 
-mod fuzz_mon_hw2;
-mod fuzz_mon_hw3;
-mod fuzz_mon_rf;
-
-mod fuzz_nav_hp_pos_llh;
-mod fuzz_nav_pvt;
+// `fuzz_*` tests are their own auto-discovered binaries; don't declare them here.
--- a/ublox/tests/common/mod.rs
+++ b/ublox/tests/common/mod.rs
@@ -1,0 +1,64 @@
+//! Shared helpers for the UBX packet fuzz tests.
+//!
+//! Keep this module minimal: only protocol-level primitives (checksum, frame
+//! envelope) and generic proptest strategies belong here. Per-packet payload
+//! structs and their `*_payload_strategy()` / `to_bytes()` helpers stay in
+//! their own test files so that each packet's correctness remains
+//! independently auditable.
+
+#![allow(dead_code)] // each test binary uses only a subset of these helpers
+
+use byteorder::{LittleEndian, WriteBytesExt};
+use proptest::prelude::*;
+use ublox::constants::{UBX_SYNC_CHAR_1, UBX_SYNC_CHAR_2};
+
+/// Calculates the 8-bit Fletcher-16 checksum used by U-Blox.
+///
+/// `data` is the checksummed region of the frame: class, id, length and
+/// payload (everything between the sync chars and the checksum bytes).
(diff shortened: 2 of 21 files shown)


andrei-ng:
@gsokoll , I'll check this one this weekend

ublox/tests/lib.rs
```diff
@@ -3,9 +3,4 @@ mod parser_binary_dump_test;
 mod parser_tests;
 mod rxm_sfrbx;
 
-mod fuzz_mon_hw2;
-mod fuzz_mon_hw3;
-mod fuzz_mon_rf;
-
-mod fuzz_nav_hp_pos_llh;
-mod fuzz_nav_pvt;
+// `fuzz_*` tests are their own auto-discovered binaries; don't declare them here.
```
CramBL, line 6: What is the motivation for this exactly?

Compiling them each as a separate test binary is really bad for compile-times.

gsokoll: Each `fuzz_*.rs` file is already its own test binary, and the `mod fuzz_*` lines in `lib.rs` pulled five of them into the lib binary which meant those five were being compiled twice. Removing the declarations dropped that redundant second compile.  Have I misunderstood this behaviour ?

That said, I agree that the per-file binaries have an impact on compile time.  This PR doesn't attempt to resolve that (it was only intended to drop the redundant second compile of those five). Merging all the fuzz tests into a single test binary sounds like a good improvement, but since it's conceptually a separate change would it perhaps be better as its own PR ?

CramBL: Right, it's a separate change.

CramBL (approved):

andrei-ng (approved):

PR #293, opened 2026-06-04
Add L5/L6 CFG-SIGNAL keys and NavIC constellation support

gsokoll:
Expands the crate's coverage to L5/L6 bands and NavIC constellations tracked by F10-generation (NEO-F10N/T) and X20-generation (ZED-X20P/D) receivers.

## Changes

_**Breaking Change:** This adds new enum variants to `GnssId` and `SigCfgMask`. Code with exhaustive pattern matches on these enums will require updates._

New `CfgVal` variants:

- SignalGpsL5Ena   (0x10310004)
- SignalGalE5aEna  (0x10310009)
- SignalGalE6Ena   (0x1031000b)  (X20-only)
- SignalBdsB1cEna  (0x1031000f)
- SignalBdsB2aEna  (0x10310028)
- SignalBdsB3Ena   (0x10310010)  (X20-only)
- SignalQzssL5Ena  (0x10310017)
- SignalNavicEna   (0x10310026)
- SignalNavicL5Ena (0x1031001d)

NavIC constellation support:

- GnssId::NAVIC = 7 (UBX gnssId for NavIC/IRNSS)
- NavicSigMask bitflags + SigCfgMask::Navic variant + dispatch arm

Proposed changes pass all tests other than semver ones (due to new enum variants mentioned above).

## Hardware verification

- The seven F10-supported keys (GPS L5, Gal E5a, BDS B1C/B2a, QZSS L5, NavIC, NavIC L5) have been tested and verified using a NEO-F10N board.
- GAL_E6_ENA and BDS_B3_ENA are X20-only. Key IDs have been sourced from the X20 ICD but not yet hardware verified.

PR #298, opened 2026-07-03
fix(tim-tm2): shift timeBase before matching in time_base()

gsokoll:
`TimTm2Flags::time_base()` masked the `timeBase` field (flags bits 3-4) with
`0b11000` but never shifted it down, so GNSS (raw 1) and UTC (raw 2) masked to
8 and 16 instead of 1 and 2. Calling `time_base()` on a TIM-TM2 message with a
GNSS or UTC time base (any time-mark once the receiver has a fix) therefore hit
`unreachable!()` and panicked; only the receiver-time case (value 0) matched, by
coincidence.

Shift the field down before matching (mirroring `tim_tp.rs::raim_active`) so all
three defined `timeBase` values decode correctly. The reserved value 3 keeps the
existing `unreachable!()` guard, so this is a non-breaking change with no public
API delta. Includes an in-module regression test covering the receiver/GNSS/UTC
bit-patterns.

### DIFF
--- a/ublox/src/ubx_packets/packets/tim_tm2.rs
+++ b/ublox/src/ubx_packets/packets/tim_tm2.rs
@@ -63,7 +63,7 @@ impl TimTm2Flags {
     }
 
     pub fn time_base(&self) -> TimTm2TimeBase {
-        match self.0 & 0b11000 {
+        match (self.0 >> 3) & 0b11 {
             0 => TimTm2TimeBase::Receiver,
             1 => TimTm2TimeBase::Gnss,
             2 => TimTm2TimeBase::Utc,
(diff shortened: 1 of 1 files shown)


gsokoll:
Note: the red `semver` jobs here are pre-existing on `master` (the NavIC `GnssId`/`SigCfgMask` additions from #293 in `cfg_gnss.rs`), not introduced by this PR.

One design note: the reserved `timeBase` value `3` still hits `unreachable!()` (unchanged behaviour). A fully panic-free version would instead return a new `TimTm2TimeBase::Reserved` variant, but that would be a breaking change, so I've kept it out of this PR. Happy to introduce the `Reserved` variant into the next breaking batch (0.11) if you'd prefer the fully panic-free API.

andrei-ng (approved):

andrei-ng:
> Note: the red `semver` jobs here are pre-existing on `master` (the NavIC `GnssId`/`SigCfgMask` additions from #293 in `cfg_gnss.rs`), not introduced by this PR.
> 
> One design note: the reserved `timeBase` value `3` still hits `unreachable!()` (unchanged behaviour). A fully panic-free version would instead return a new `TimTm2TimeBase::Reserved` variant, but that would be a breaking change, so I've kept it out of this PR. Happy to introduce the `Reserved` variant into the next breaking batch (0.11) if you'd prefer the fully panic-free API.

Hi @gsokoll ,
Yes, I think that would be better. 
The smever is already violated, so better to do all other breaking changes now before a version bump.

I merged the PR already, so if you could do a small PR with that additional fix that would be awesome

